PR #56, opened 2018-07-28
Capture the DX, VPN and Peering connections for all VPCs

jchrisfarris:
Please pay extra attention to both how I structured this json and the fact I'm putting it in the unused VPC section of the database.


SpenGietz (changes requested):
A few changes requested. Solid module 👍

SpenGietz (approved):

PR #66, opened 2018-08-06
Feature/enum permissions bruteforce

alexanderinsa:


BenFriedland-Rhino (approved):
The enum_permissions_bruteforce module could use a refactor in the future, but it's currently functional.

Issue #74, opened 2018-08-22
AccessDeniedException not handled throughout enum_monitoring module


ansorren:
Using a key that doesn't have permissions to make those calls will result in the AccessDeniedException, and since it's not passed in the module the enumeration will stop.

For example:
  <class 'botocore.exceptions.ClientError'>: An error occurred (AccessDeniedException) when calling the ListDetectors operation: User: arn:aws:iam::xxx:user/janedoe is not authorized to perform: guardduty:ListDetectors on resource: arn:aws:guardduty:ap-southeast-1:xxx:detector/*

SpenGietz:
This fix has been merged into dev: https://github.com/RhinoSecurityLabs/pacu/pull/75

It will be merged into master next Monday or Tuesday I believe. Thanks for your report! I'll close this when it makes it to master

SpenGietz:
Just now merged to master!

PR #75, opened 2018-08-22
Added permission checking

alexanderinsa:
PR to resolve https://github.com/RhinoSecurityLabs/pacu/issues/74.

### DIFF
--- a/modules/enum_monitoring/main.py
+++ b/modules/enum_monitoring/main.py
@@ -40,6 +40,73 @@ parser.add_argument('--config', required=False, default=False, action='store_tru
 parser.add_argument('--vpc', required=False, default=False, action='store_true', help='Enumerate VPC flow logs.')
 
 
+def permission_check(pacu, args):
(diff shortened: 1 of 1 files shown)


modules/enum_monitoring/main.py
```diff
@@ -40,6 +40,73 @@
...
+    if args.guard_duty or enum_all:
+        try:
+            client = pacu.get_boto3_client('guardduty', 'us-east-1')
+            client.list_detectors()
+            confirmed.append('guardduty')
+        except ClientError as error:
+            code = error.response['Error']['Code']
+            if code != 'AccessDeniedException':
+                pacu.print(code)
+    if args.config or enum_all:
```
SpenGietz: This doesn't hit all the API calls that Config makes. It might be a better idea to just wrap the calls in the main function and if there is an error thrown, remove that group from the targets for each for loop iteration

SpenGietz (changes requested):

SpenGietz (approved):

PR #76, opened 2018-08-22
Module/ebs snapshot explorer

alexanderinsa:


modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
+import argparse
+
+module_info = {
+    'name': 'ebs_snapshot_explorer',
+    'author': 'Alexander Morgenstern alexander.morgenstern@rhinosecuritylab.com',
```
SpenGietz: rhinosecuritylabS.com

Is this in any other modules?

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
+import argparse
```
SpenGietz: Is this supposed to be in __init__.py? If we are going to make changes like this, we should standardize them if possible

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+module_info = {
+    'name': 'ebs_snapshot_explorer',
+    'author': 'Alexander Morgenstern alexander.morgenstern@rhinosecuritylab.com',
+    'category': 'post_exploitation',
+    'one_liner': 'Loads EBS snapshots and volumes so they are easily accessible.',
```
SpenGietz: This doesn't really make sense without context

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    'description': 'This module will use an EC2 instance to load Elastic Block Store volumes and snapshots in the account and allow the user to access the data.',
```
SpenGietz: Do something like

"This module will cycle through existing EBS volumes and create snapshots of them, then restore those snapshots and existing snapshots to new EBS volumes, which will then be attached to the supplied EC2 instance for you to mount. This will give you access to the files on the various volumes, where you can then look for sensitive information. Afterwards, it will cleanup the created volumes and snapshots by detaching them from your instance and removing them from the AWS account."

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    'services': ['EC2'],
+    'prerequisite_modules': [],
+    'arguments_to_autocomplete': ['--instance', '--region', '--zone'],
```
SpenGietz: --instance-id

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+}
+
+parser = argparse.ArgumentParser(add_help=False, description=module_info['description'])
+parser.add_argument(
+    '--instance',
```
SpenGietz: --instance-id

modules/ebs_snapshot_explorer/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    required=True,
+    default=None,
```
SpenGietz: For this and the other two, you don't need/want a default if it is required, as you will never have a default state

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+
+from botocore.exceptions import ClientError
+
+from . import parser
+from . import module_info
+
+
+def input_helper(input):
+    """Helper function that loops until a successful response is given"""
+    prompt = '    Load next set of volumes? (y/n) '
```
SpenGietz: Probably could just put this into the input, since this is the only place it is being used

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    # load volume set
+    set_index = 0
+    set_count = 10
+
+    while set_index < len(volume_ids):
+        current_volume_set = volume_ids[set_index:set_index + set_count]
+        waiter = client.get_waiter('volume_available')
+        waiter.wait(VolumeIds=current_volume_set)
+        attached = modify_volume_set(
+            client, print, 'attach_volume', instance_id, current_volume_set)
```
SpenGietz: move the ) to the next line, kind of confusing with this single line break

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        if not attached:
+            print(' Volume attachment failed')
+            print(' Exiting...')
+            return False
```
SpenGietz: So if a single attachment fails, the whole loop exits? What happens to the stale volumes and everything?

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        running = input_helper(input)
+        detached = modify_volume_set(
+            client, print, 'detach_volume', instance_id, current_volume_set)
```
SpenGietz: Same thing with the )

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+            return False
+        waiter.wait(VolumeIds=current_volume_set)
+        set_index += set_count
+        if not running:
+            break
+    cleanup(client)
+    return True
+
+
+def modify_volume_set(client, print, func, instance_id, volume_id_set):
```
SpenGietz: volume_id_set here, volume_ids on line 74

A set is a python type, so try to avoid having "set" in the name if it is a list

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        except ClientError as error:
+            code = error.response['Error']['Code']
+            if  code == 'UnauthorizedOperation':
+                print('  FAILURE MISSING AWS PERMISSIONS')
+            else:
+                print(error)
+            return False
+    return True
+
+def get_valid_device(client, instance):
```
SpenGietz: Just for clarity, it would probably make more sense to keep the name of this variable the same across different functions (instance_id instead of instance)

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    """Returns the next device mapping available
+
+    Args:
+        client (boto3.client): Client that gets the current block device mappings
+        instance (str): InstanceId to get curretn block device mappings
+    Returns:
+        str: Returns next mapping in form of /dev/xvd[base], otherwise /dev/xvdzz
+
+    """
+    response = client.describe_instances(InstanceIds=[instance])
```
SpenGietz: Ideally, this would check the Pacu database to see if this instance has already been enumerated so we don't have to waste an API call everytime this is invoked

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    mappings = response['Reservations'][0]['Instances'][0]['BlockDeviceMappings']
+    current_mappings = [device['DeviceName'] for device in mappings]
+    base_mappings = [char for char in 'bcdefghijklmnoqrstuvwxyz']
```
SpenGietz: Is this just a list of those letters?

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    for base in base_mappings:
+        if '/dev/xvd' + base not in current_mappings:
+            return '/dev/xvd' + base
+    return '/dev/xvdzz'
```
SpenGietz: What if this is in there already?

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def get_snapshots(pacu, session, region):
```
SpenGietz: Is this supposed to be region or regions? On line 122 it says it is a list of regions

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        session (PacuSession): Reference to the Pacu session database
+        regions (list): Regions to check for snapshots
+    Returns:
+        dict: Mapping regions to corresponding list of snapshot_ids.
+    """
+    ec2_data = deepcopy(session.EC2)
+    if 'Snapshots' not in ec2_data:
+        fields = ['EC2', 'Snapshots']
+        module = 'enum_ebs_volumes_snapshots'
+        fetched_ec2_instances = pacu.fetch_data(fields, module)
```
SpenGietz: The name `fetched_ec2_instances` seems misplaced here

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        if fetched_ec2_instances is False:
+            return None
+    snapshot_ids = []
+    for snapshot in ec2_data['Snapshots']:
```
SpenGietz: I think that if ec2_data exists but Snapshots isn't in it, ec2_data won't be updated by the fetch_data call because you deepcopied this dict before fetching the data and didn't replace it

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        if snapshot['Region'] == region:
+            snapshot_ids.append(snapshot['SnapshotId'])
+    return snapshot_ids
+
+
+def get_volumes(pacu, session, region):
```
SpenGietz: Same question for `region` here. Looks like it is supposed to be `region` and not `regions` like on line 145

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        session (PacuSession): Reference to the Pacu session database
+        regions (list): Regions to check for volumes
+    Returns:
+        dict: Mapping regions to corresponding list of volume_ids.
+    """
+    ec2_data = deepcopy(session.EC2)
+    if 'Volumes' not in ec2_data:
+        fields = ['EC2', 'Volumes']
+        module = 'enum_ebs_volumes_snapshots'
+        fetched_ec2_instances = pacu.fetch_data(fields, module)
```
SpenGietz: `fetched_ec2_instances` name again

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+        if fetched_ec2_instances is False:
+            return None
+    volume_ids = []
+    for volume in ec2_data['Volumes']:
```
SpenGietz: Same problem with fetch_data

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def generate_volumes_from_snapshots(client, snapshots, zone):
+    """Returns a list of generated volumes"""
+    volume_ids = []
+    waiter = client.get_waiter('snapshot_completed')
+    waiter.wait(SnapshotIds=snapshots)
+    for snapshot in snapshots:
+        response = client.create_volume(
+            SnapshotId=snapshot, AvailabilityZone=zone)
```
SpenGietz: Confusing line break with the )

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    return failed_snapshots
+
+
+def cleanup(client):
+    """Cleans up the temporary snapshots and volumes created during this
+    modules execution
+    """
+    new_data = {}
+    success = True
+    temp_file = Path(__file__).parent / 'temp.json'
```
SpenGietz: Tested this and it seems to work, but seems like it should be trying to divide with `/` instead of concatting just by the looks of it

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+
+def main(args, pacu):
+    """Main module function, called from Pacu"""
+    args = parser.parse_args(args)
+    session = pacu.get_active_session()
+    print = pacu.print
+
+    instance = args.instance
+    region = args.region
+    zone = region + args.zone
```
SpenGietz: This should be a little more clear in the args. I was thinking of zone as `us-east-1a` instead of just `a`

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    volumes = get_volumes(pacu, session, region)
+    summary_data = {'snapshots': len(snapshots), 'volumes': len(volumes)}
+
+    print('  Attaching volumes...')
+    temp_snaps = generate_snapshots_from_volumes(client, volumes)
+    temp_volumes = generate_volumes_from_snapshots(client, temp_snaps, zone)
+    load_volumes(client, pacu.print, pacu.input, instance, temp_volumes)
+    print('  Finished attaching volumes')
+
+    print('  Attaching volumes from existing snapshots')
```
SpenGietz: Consistency with the `...` from line 280

modules/ebs_snapshot_explorer/main.py
```diff
@@ -0,0 +1,303 @@
...
+    temp_volumes = generate_volumes_from_snapshots(client, snapshots, zone)
+    load_volumes(client, pacu.print, pacu.input, instance, temp_volumes)
+    print('  Finished attaching existing snapshot volumes ')
+
+    return summary_data
+
+
+def summary(data, pacu):
```
SpenGietz: Looks like `pacu` doesn't need to get passed in here

SpenGietz (changes requested):

modules/ebs_snapshot_explorer/main.py
```diff
@@ -1,26 +1,38 @@
...
-            return True
-        elif response.lower() == 'n':
-            return False
-
-def load_volumes(client, print, input, instance_id, volume_ids):
+module_info = {
+    'name': 'ebs_snapshot_explorer',
+    'author': 'Alexander Morgenstern alexander.morgenstern@rhinosecuritylabs.com',
+    'category': 'post_exploitation',
+    'one_liner': 'Loads EBS snapshots and volumes so they are easily accessible.',
```
SpenGietz: I think something like: "Restores and attaches EBS volumes/snapshots to an EC2 instance of your choice."

modules/ebs_snapshot_explorer/main.py
```diff
@@ -104,40 +128,68 @@ def get_valid_device(client, instance):
...
     Returns:
-        dict: Mapping regions to corresponding list of snapshot_ids.
+        list: List of Instances.
     """
-    ec2_data = deepcopy(session.EC2)
+    ec2_data = deepcopy(pacu.get_active_session().EC2)
+    if 'Instances' not in ec2_data:
+        fields = ['EC2', 'Instances']
+        module = 'enum_ec2'
+        fetched_ec2_instances = pacu.fetch_data(fields, module)
```
SpenGietz: Make sure to pass `--instances` to enum_ec2 so that you don't run the full enum_ec2 module

modules/ebs_snapshot_explorer/main.py
```diff
@@ -104,40 +128,68 @@ def get_valid_device(client, instance):
...
+        if fetched_ec2_instances is False:
+            return []
+    return ec2_data['Instances']
```
SpenGietz: I still think that pacu.fetch_data won't get any results into this spot if it enumerates new ones

modules/ebs_snapshot_explorer/main.py
```diff
@@ -104,40 +128,68 @@ def get_valid_device(client, instance):
...
+        pacu (Main): Reference to Pacu
+    Returns:
+        list: List of Snapshots.
+    """
+    ec2_data = deepcopy(pacu.get_active_session().EC2)
     if 'Snapshots' not in ec2_data:
         fields = ['EC2', 'Snapshots']
         module = 'enum_ebs_volumes_snapshots'
-        fetched_ec2_instances = pacu.fetch_data(fields, module)
+        fetched_snapshots = pacu.fetch_data(fields, module)
```
SpenGietz: Pass in `--snapshots` to only enumerate snapshots here instead of everything

modules/ebs_snapshot_explorer/main.py
```diff
@@ -104,40 +128,68 @@ def get_valid_device(client, instance):
...
         if fetched_ec2_instances is False:
-            return None
-    snapshot_ids = []
-    for snapshot in ec2_data['Snapshots']:
-        if snapshot['Region'] == region:
-            snapshot_ids.append(snapshot['SnapshotId'])
-    return snapshot_ids
+            return []
+    return ec2_data['Snapshots']
```
SpenGietz: Same as above, I don't think pacu.fetch_data will fill this up

modules/ebs_snapshot_explorer/main.py
```diff
@@ -146,18 +198,14 @@ def get_volumes(pacu, session, region):
     Returns:
         dict: Mapping regions to corresponding list of volume_ids.
     """
-    ec2_data = deepcopy(session.EC2)
+    ec2_data = deepcopy(pacu.get_active_session().EC2)
     if 'Volumes' not in ec2_data:
         fields = ['EC2', 'Volumes']
         module = 'enum_ebs_volumes_snapshots'
-        fetched_ec2_instances = pacu.fetch_data(fields, module)
+        fetched_volumes = pacu.fetch_data(fields, module)
```
SpenGietz: Pass in `--volumes`

modules/ebs_snapshot_explorer/main.py
```diff
@@ -146,18 +198,14 @@ def get_volumes(pacu, session, region):
...
         if fetched_ec2_instances is False:
-            return None
-    volume_ids = []
-    for volume in ec2_data['Volumes']:
-        if volume['Region'] == region:
-            volume_ids.append(volume['VolumeId'])
-    return volume_ids
+            return []
+    return ec2_data['Volumes']
```
SpenGietz: same problem as above with pacu.fetch_data

SpenGietz (changes requested):

modules/ebs_snapshot_explorer/main.py
alexanderinsa: These are the required arguments for the summary function. Could set the default value in Pacu to None so that it's optional.

modules/ebs_snapshot_explorer/main.py
alexanderinsa: I'm not sure how it works, it might be operator overload for Path objects, but in the official docs, it looks like that's the proper way to append parts to a Path object.

SpenGietz (approved):

PR #77, opened 2018-08-24
Feature/condense output

alexanderinsa:


modules/enum_glue/main.py
```diff
@@ -45,6 +47,24 @@
...
+        data = response[key]
+        while 'NextToken' in response and response['NextToken'] != '':
+            print({**kwargs, **{'NextToken': response['NextToken']}})
+            response = caller({**kwargs, **{'NextToken': response['NextToken']}})
+            data.extend(response[key])
+        return data
+    except ClientError as error:
+        code = error.response['Error']['Code']
+        if code == 'AccessDeniedException':
+            print('  FAILURE: MISSING NEEDED PERMISSIONS')
```
SpenGietz: Will the user be able to tell which API call this references?

modules/enum_glue/main.py
```diff
@@ -45,6 +47,24 @@
...
 parser.add_argument('--jobs', required=False, default=False, action='store_true', help='Enumerate Glue jobs.')
 
+def fetch_glue_data(client, func, key, print, **kwargs):
+    caller = getattr(client, func)
+    try:
+        response = caller(**kwargs)
+        data = response[key]
+        while 'NextToken' in response and response['NextToken'] != '':
+            print({**kwargs, **{'NextToken': response['NextToken']}})
+            response = caller({**kwargs, **{'NextToken': response['NextToken']}})
```
SpenGietz: I think the 'Region' arg needs to get input here for these items (like in some of the code removed below)

modules/enum_glue/main.py
```diff
@@ -73,144 +93,39 @@ def main(args, pacu_main):
...
-                    connection['Region'] = region
-                    connections.append(connection)
-
-                while 'NextToken' in response:
-                    response = client.get_connections(
-                        MaxResults=200,
-                        NextToken=response['NextToken']
-                    )
-                    for connection in response['ConnectionList']:
-                        connection['Region'] = region
```
SpenGietz: This is the Region line I was refering to above

modules/enum_lambda/main.py
```diff
@@ -37,10 +37,9 @@
 parser.add_argument('--regions', required=False, default=None, help='One or more (comma separated) AWS regions in the format us-east-1. Defaults to all session regions.')
 
 
-def fetch_lambda_data(client, func, key, **kwargs):
+def fetch_lambda_data(client, func, key, print, **kwargs):
     caller = getattr(client, func)
     try:
-        print('  Starting enumeration of {}'.format(key))
         response = caller(**kwargs)
         data = response[key]
```
SpenGietz: Is the lambda region distinguishable by region in some way?

modules/enum_lightsail/main.py
```diff
@@ -63,22 +63,21 @@ def camelCase(name):
...
+def fetch_lightsail_data(client, func, print):
     # Adding 'get_' portion to each field to build command.
     caller = getattr(client, 'get_' + func)
-    print('  Attempting to enumerate {}'.format(func))
     try:
         response = caller()
         data = response[camelCase(func)]
         while 'nextPageToken' in response:
             response = caller(pageToken=response['nextPageToken'])
             data.extend(response[camelCase(func)])
```
SpenGietz: Are these separated by region in the DB somehow? or identifiable at least

modules/enum_monitoring/main.py
```diff
@@ -266,11 +266,11 @@ def main(args, pacu_main):
         cw_data = deepcopy(session.CloudWatch)
         cw_data['Alarms'] = all_alarms
         session.update(pacu_main.database, CloudWatch=cw_data)
-        print('  {} total CloudWatch alarms found.\n'.format(len(session.CloudWatch['Alarms'])))
+        print('  {} total CloudWatch alarms found.'.format(len(session.CloudWatch['Alarms'])))
```
SpenGietz: remove "found" to be consistent

modules/generate_temp_lightsail_access/main.py
```diff
@@ -126,21 +126,21 @@ def main(args, pacu_main):
...
+                print('    Successfully created temporary access for {}'.format(name))
             except ClientError as error:
                 code = error.response['Error']['Code']
                 if code == 'AccessDeniedException':
-                    print('Unauthorized to generate temporary access.')
+                    print('     Unauthorized to generate temporary access.')
                     return
                 elif code == 'OperationFailureException':
-                    print(error.response['Error']['Message'])
+                    print(error.response['Error']['Code'])
```
SpenGietz: Is this just printing 'OperationFailureException' everytime? Why is it caught explicitly?

SpenGietz (changes requested):

modules/create_api_gateway_keys/main.py
```diff
@@ -106,14 +110,15 @@ def main(args, pacu_main):
...
 
 
 def summary(data, pacu_main):
     out = ''
     if data.get('cleanup'):
-        out += '  Old keys were removed.\n'
-    else:
-        out += '  Old keys were not removed.\n'
-    out += '  {} key(s) were created.\n'.format(data['keys_created'])
+        out += '  Old keys removed.\n'
```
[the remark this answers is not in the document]
alexanderinsa: It could, I guess the only reason is you can cleanup your keys and then generate a new set. How should I go into more detail about how the keys are stored in the database? Tell them to look in APIGateway column for the current session?

modules/download_lightsail_ssh_keys/main.py
```diff
@@ -34,29 +33,47 @@
...
         client = pacu_main.get_boto3_client('lightsail', region)
         downloaded_keys = client.download_default_key_pair()
         restructured_keys = {
             'publicKey': downloaded_keys['publicKeyBase64'],
             'privateKey': downloaded_keys['privateKeyBase64']
         }
-        print('Region: {}\n{}\n'.format(region, json.dumps(restructured_keys)))
+
+        private_path = cur_path / 'default'
+        with private_path.open('w', encoding='utf-8') as key_file:
```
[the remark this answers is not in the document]
alexanderinsa: I set the flag to 'w' because I just wanted to write the files. I didn't want there to be a chance that the old data persisted in that key-file because the latest key downloaded should be the most correct.

modules/enum_account/main.py
```diff
@@ -95,8 +95,6 @@ def main(args, pacu_main):
 
     session.update(pacu_main.database, Account=account_data)
 
-    print('{} completed.\n'.format(module_info['name']))
```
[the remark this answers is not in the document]
alexanderinsa: Took out the line because Ben made a comment that the Module Summary by itself was concise enough for the whole module. I can add it back in though. I added the print statement because it explicitly lets the user know the account that is being enumerated against.

modules/generate_temp_lightsail_access/main.py
alexanderinsa: OperationFailureException generally occurs when you try to generate temporary access for a non running instance. I can be more explicity in the failure message.

modules/download_ec2_userdata/main.py
```diff
@@ -1,119 +1,119 @@
...
+    args = parser.parse_args(args)
+    print = pacu_main.print
+    fetch_data = pacu_main.fetch_data
+    ######
+
+    instances = []
+    summary_data = {'userdata_downloads': 0}
+    # Check permissions before doing anything
+    try:
+        client = pacu_main.get_boto3_client('ec2', 'us-east-1')
```
SpenGietz: This should get changed to be a region from get_regions() incase us-east-1 isn't in the session regions

modules/enum_ec2/main.py
```diff
@@ -1,483 +1,482 @@
...
+def summary(data, pacu_main):
+    results = []
+
+    results.append('  Regions:')
+    for region in data['regions']:
+        results.append('     {}'.format(region))
+
+    results.append('')
+
+    if len(data['Instances']) > 0:
```
SpenGietz: This is a hard problem to solve with how summary is setup but we still want to show when 0 of something was found, we only want to hide things that weren't even attempted for enumerartion

modules/enum_elb_logging/main.py
```diff
@@ -1,117 +1,119 @@
...
+
+
+    print('\n{} completed.\n'.format(module_info['name']))
+    return summary_data
+
+
+def summary(data, pacu_main):
+    out = '  {} Load balancer(s) have been found\n'.format(data['load_balancers'])
+    if data['load_balancers'] > 0:
+        out += '  List of Load balancers without logging saved to:\n    {}\n'.format(data['csv_file_path'])
```
SpenGietz: Maybe add a line saying how many didn't have logging enabled

modules/enum_glue/main.py
```diff
@@ -1,154 +1,156 @@
...
+    session.update(pacu_main.database, Glue=glue_data)
+
+    print('\n{} completed.\n'.format(module_info['name']))
+    return summary_data
+
+
+def summary(data, pacu_main):
+    out = ''
+    for key in data:
+        out += '  {} total {}(s) found.\n'.format(data[key], key[:-1])
```
SpenGietz: ideally we only print what was enumerated, so if --connections was passed in, only show enumerated connections, even if it is 0

modules/enum_lambda/main.py
```diff
@@ -1,139 +1,140 @@
...
+    'services': ['Lambda'],
+
+    # For prerequisite modules, try and see if any existing modules return the data that is required for your module before writing that code yourself, that way, session data can stay separated and modular.
+    'prerequisite_modules': [],
+
+    # External resources that the module depends on. Valid options are either a GitHub URL (must end in .git) or single file URL.
+    'external_dependencies': [],
+
+    # Module arguments to autocomplete when the user hits tab
+    'arguments_to_autocomplete': ['--versions-all'],
```
SpenGietz: Missing the --regions argument here

modules/enum_lightsail/main.py
```diff
@@ -1,125 +1,125 @@
...
+    print('{} completed.\n'.format(module_info['name']))
+    return summary_data
+
+
+def summary(data, pacu_main):
+    out = '  Regions Enumerated:\n'
+    for region in data['regions']:
+        out += '    {}\n'.format(region)
+    del data['regions']
+    for field in data:
```
SpenGietz: Same as the other modules, we should only show results for what was enumerated, even if there is 0 of something

modules/enum_monitoring/main.py
```diff
@@ -1,351 +1,351 @@
...
+        print('  {} total VPC flow logs'.format(len(session.VPC['FlowLogs'])))
+        summary_data['flowlogs'] = len(all_flow_logs)
+
+    print('\n{} completed.\n'.format(module_info['name']))
+    return summary_data
+
+
+def summary(data, pacu_main):
+    out = ''
+    if 'ShieldSubscription' in data:
```
SpenGietz: Same as the other modules, we still should summarize results that found nothing, we only want to hide ones that weren't even enumerated

modules/generate_temp_lightsail_access/main.py
```diff
@@ -1,172 +1,172 @@
...
+                )
+                temp_keys[region].append(response['accessDetails'])
+                print('    Successfully created temporary access for {}'.format(name))
+            except ClientError as error:
+                code = error.response['Error']['Code']
+                if code == 'AccessDeniedException':
+                    print('      Unauthorized to generate temporary access.')
+                    return
+                elif code == 'OperationFailureException':
+                    print('      FAILED: Instance Stopped')
```
SpenGietz: It failed because the instance is offline? A little unclear here

SpenGietz (changes requested):

modules/enum_monitoring/main.py
alexanderinsa: This should be working as intended, the summary_data fields are only updated if the field is actually enumerated.

modules/cloudtrail_csv_injection/main.py
```diff
@@ -89,25 +82,29 @@ def main(args, pacu_main):
...
     print('\n{} completed.\n'.format(module_info['name']))
     return summary_data
 
 
 def summary(data, pacu_main):
-    out += '  {} trail(s) attacked.\n'.format(data['trail_attacks'])
-    out += '  {} trail(s) attack fails.\n'.format(data['trail_fails'])
-    out += '  {} instances() attacked.\n'.format(data['instance_attacks'])
-    out += '  {} instance(s) attack fails.'.format(data['instance_fails'])
+    out = '  {} trail(s) successfully attacked.\n'.format(data['success'])
```
SpenGietz: Change this and the next line to "{} CloudTrail regions" rather than trails

SpenGietz (changes requested):

SpenGietz (dismissed):

PR #87, opened 2018-09-05
Renamed s3 module

alexanderinsa:


modules/dl_s3_bucket/main.py
```diff
@@ -9,7 +9,7 @@
 
 module_info = {
     # Name of the module (should be the same as the filename)
-    'name': 's3_bucket_dump',
+    'name': 'dl_s3_bucket',
```
SpenGietz: There are other places in the module with the old name. Change to module_info['name'] so we don't have to worry about that in the future

SpenGietz (changes requested):

SpenGietz (approved):

PR #100, opened 2018-09-20
Feature/add completion to core

alexanderinsa:


modules/lambda__enum/main.py
```diff
@@ -43,15 +43,13 @@ def fetch_lambda_data(client, func, key, print, **kwargs):
...
-            print('    Found {} data'.format(key))
             return data
         while 'nextMarker' in response:
             response = caller({**kwargs, **{'NextMarker': response['nextMarker']}})
             data.extend(response[key])
-        print('    Found {} {}'.format(len(data), key))
         return data
     except client.exceptions.ResourceNotFoundException:
-        print('    No valid {} found'.format(key))
+        pass
```
SpenGietz: When would this error come up? Seems like we shouldn't just pass on it

SpenGietz (changes requested):

modules/lambda__enum/main.py
alexanderinsa: That exception is raised if you try to get the policy attached to a Lambda function and it doesn't have one. I decided to take out the print statement was because it seems like evidence enough that the field is not populated in the database that there was no policy associated with that Lambda function.

SpenGietz (approved):

PR #115, opened 2018-10-13
pin urllib3 version to avoid requests incompatible error.

mayn:
1. change requirements.txt encoding to utf-8. It was previously utf-16
```
file -I pacu/requirements.txt
pacu/requirements.txt: text/plain; charset=utf-16le
```

2. pint urllib3 version due to requests incompatibility.
The following error was being displayed during installation. (the installation script would still exit 0)
screen shot 2018-10-13 at 2 01 12 pm

after pinning error went away.


SpenGietz (approved):
Thank you for the fixes! Merged into Dev now and I'll be merging it into master soon

PR #126, opened 2018-11-26
Bug/fix python regex

jaywon:
Had some issues with install script in that Regex didn't match additional string output in the case of using Anaconda version of Python.

Also, had issue with `$(python3 --version)` output not being in `PYTHON_VERSION` variable, redirected stderr and stdout and that fixed that issue as well.

screen shot 2018-11-25 at 8 02 13 pm

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ check_python_version() {
     echo "$PYTHON_VERSION"
 
     # This ensures the Python version is 3.6 or higher. https://regex101.com/
-    VERSION_REGEX="Python\ ((3\.[^0-4])|(3\.[1-9][0-9]+)|([4-9]+\.\d+))"
+    VERSION_REGEX="Python\ ((3\.[^0-4])|(3\.[1-9][0-9]+)|([4-9]+\.\d+)).*"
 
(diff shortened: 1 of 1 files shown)


jukefr:
Weird, I don't have this problem with Anaconda on macOs

```bash
$ which python
/usr/local/anaconda3/bin/python
```

install.sh ran without any issues

jaywon:
Hmmm...I'm on the same setup. 

screen shot 2018-11-28 at 1 42 25 pm

jaywon:
The issue i thought originally was the regex but what I tracked down was that the `PYTHON_VERSION` var wasn't receiving the output of the command substitution. It was going to `stdout` but the variable would be empty causing the failure to detect version. Already had the additional regex so left in PR.

SpenGietz:
Hey @jaywon thanks for the PR and bug report. I'm sorry about the delayed response here.

@BenFriedland-Rhino would you mind looking into this? Looks like the problem just has to do with the PYTHON_VERSION variable like stated in the comment before this.

BenFriedland-Rhino (approved):

PR #129, opened 2019-01-02
Correcting comment about Python version regex in install.sh

Spencer-Doak:
Just correcting the comment about the Python version matching RegEx pattern. (The pattern checks for at least 3.5, but the prior comment stated: "This ensures the Python version is 3.6 or higher.")


SpenGietz:
Good catch, thank you!

PR #142, opened 2019-06-24
enhancement: added the ability to load commands from a file

manasmbellani:
Added the ability to run a list of commands from a commands file, which can be loaded via command 'load_commands_file'

### DIFF
--- a/pacu.py
+++ b/pacu.py
@@ -1171,7 +1171,7 @@ class Main:
         elif command_name == 'exit' or command_name == 'quit':
             print('\n    exit/quit\n        Exit Pacu\n')
         elif command_name == 'load_commands_file':
-            print('\n    load_commands_file\n <commands_file>        Load an existing file with a set of commands to execute')
+            print('\n    load_commands_file <commands_file>\n        Load an existing file with a set of commands to execute')
         else:
             print('Command or module not found. Is it spelled correctly? Try using the module search function.')
         return
--- a/pacu.py
+++ b/pacu.py
@@ -552,7 +552,8 @@ class Main:
             for command in commands:
                 print("Executing command: {} ...".format(command))
                 command_without_space = command.strip()
-                self.parse_command(command_without_space)
+                if command_without_space:
+                    self.parse_command(command_without_space)
             
     def parse_awscli_keys_import(self, command):
         if len(command) == 1:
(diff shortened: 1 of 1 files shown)


pacu.py
```diff
@@ -1148,6 +1170,8 @@ def display_command_help(self, command_name):
             print('\n    swap_keys\n        Change the currently active AWS key to another key that has previously been set for this session\n')
         elif command_name == 'exit' or command_name == 'quit':
             print('\n    exit/quit\n        Exit Pacu\n')
+        elif command_name == 'load_commands_file':
+            print('\n    load_commands_file\n <commands_file>        Load an existing file with a set of commands to execute')
```
SpenGietz: Should this line be
```
print('\n    load_commands_file <commands_file>\n        Load an existing file with a set of commands to execute')
```
instead? That way the "<commands_file>" stays on the same line as the load_commands_file

SpenGietz (changes requested):
Awesome PR and thanks for submitting! I only added a single comment, everything else looks great. I'll try to do some testing soon and then we can get it merged into master!

pacu.py
manasmbellani: Hi @SpenGietz 
Yes, you are right, it should be the way you have described it. 
Would you like me to make this change and push the commit into this PR again?
Thanks.

SpenGietz: Sure that would be great!

manasmbellani: Ok - I have pushed a new commit to fix the newline in help message. Could we merge this PR?
Thanks.

SpenGietz:
Sorry about the delay on this, just did some testing and this is awesome!! I can't believe how useful this will be and how simple it was to implement! Many thanks, I'll tweet out an announcement for this as well.

SpenGietz:
You should join the Pacu-CloudGoat Slack team if you're interested. We're still working on building it, but hoping to do some pre-releases for Pacu/CloudGoat/AWS security in general in there and get some good discussion going.

Here is an invite link:

https://join.slack.com/t/pacu-cloudgoat/shared_invite/enQtNDE3OTk0MjA3NTA2LTRmOTVmZjEyYjIzOTYxMGJmZDc4ZDVkOGU3ZmJlOWZhNzdkYWQ2ZmQxNTFjZThjMmJlMDFmMTU4NzUwMDM2NmY

PR #150, opened 2019-11-22
Bug fix: false positive match in case of wildcard permissions.

fruh:
There was a false positive match for PE in case of following permission e.g.:

'ec2:*VpcPeeringConnection' would match 'ec2:RunInstances'

Because the split('*') returns 'ec2:', where 'ec2:RunInstances' starts with 'ec2:'


SpenGietz:
Very helpful, thanks! Surprised it was such a simple change to get that working

PR #154, opened 2020-03-19
Fixing aws__enum_spend when account uses too many services

PenelopeFudd:
When there's more than about 40 services used, the output of the aws__enum_spend module can exceed the pacu-defined 1000 character limit and die.  

This patch raises the limit to 10,000 characters, formats the output into columns, and sorts the output by amount spent in descending order.


SpenGietz:
Thank you, great addition!

PR #156, opened 2020-04-12
Hotfix/aws enum account error handling

manasmbellani:
This handles the ClientError to ensure that even if aws account alias cannot be acquired, the rest of the script can continue.


SpenGietz:
Thank you for the PR, great addition!

PR #176, opened 2020-05-14
Fixed keyError

Iorpim:
Clearing the response object raises an exception, as it is used in the next function call, client.list_builds(...). As it is reassigned in the subsequent line, just removing it is the simplest, and best, solution.


seb1055:
Thanks for the Pull Request! Can you please share how to recreate this exception?

PR #187, opened 2020-06-19
Module/rds__enum_snapshots

sgn00:
This is a simple module that is heavily based on the ebs__enum_volumes_snapshots module. It will allow enumeration of rds snapshots in the account, as well as snapshots shared with the account by other accounts. It will check for encryption and can check for snapshot permissions.

Sample output:
image

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -81,6 +81,7 @@ class PacuSession(Base, ModelUpdateMixin):
         'Inspector',
         'Lambda',
         'Lightsail',
+        'RDS',
         'S3',
         'SecretsManager',
         'Shield',
@@ -119,6 +120,7 @@ class PacuSession(Base, ModelUpdateMixin):
     Inspector = Column(JSONType, nullable=False, default=dict)
     Lambda = Column(JSONType, nullable=False, default=dict)
     Lightsail = Column(JSONType, nullable=False, default=dict)
+    RDS = Column(JSONType, nullable=False, default=dict)
     S3 = Column(JSONType, nullable=False, default=dict)
     SecretsManager = Column(JSONType, nullable=False, default=dict)
     SSM = Column(JSONType, nullable=False, default=dict)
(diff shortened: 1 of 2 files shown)


seb1055:
Thank you for your contribution, I will be testing this module soon!

core/models.py
```diff
@@ -119,6 +120,7 @@ class PacuSession(Base, ModelUpdateMixin):
     Inspector = Column(JSONType, nullable=False, default=dict)
     Lambda = Column(JSONType, nullable=False, default=dict)
     Lightsail = Column(JSONType, nullable=False, default=dict)
+    RDS = Column(JSONType, nullable=False, default=dict)
```
seb1055, line 123: This is done correctly. Updating the models is a breaking change. This will need to be taken into consideration when merging as it will invalidate all existing db files.

modules/rds__enum_snapshots/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # Start RDS Snapshots in this region
+        count = 0
+        response = None
+        next_token = False
+
+        while (response is None or 'NextToken' in response):
+            if next_token is False:
+                try:
+                    response = client.describe_db_snapshots(
```
seb1055, line 109: Calling this method only returns snapshots of DB instances. There also exists snapshot of DB clusters. It would be great if we update this to include DB cluster snapshots as well! Here is a link to the method used to get that data. https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-cluster-snapshots.html

seb1055, line 109: This is good although it won't return db snapshots of AWS Aurora DB's as they are considered clusters. To make sure we include Aurora DB's we will need to also add [describe-db-cluster-snapshots](https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-cluster-snapshots.html)

sgn00:
I have added support for rds cluster snapshots. I made some changes to the code flow to support this, and for better error handling and print output. Do let me know if there are other necessary changes. Thanks!

seb1055:
Hi @sgn00 I wanted to give you an update on this. The code has been reviewed and agreed upon as an addition that we would like to add. We are still working on a timeline for when to merge as it will introduce a breaking change in the DB file. 

Thanks again!

RyanJarv (approved):
Sorry for the delay here. We'll go ahead and merge this into the dev branch and get a release out on master later this week.

PR #193, opened 2020-07-23
Added new module guardduty__list_findings, and fixed bug in detection__enum_services when enum GuardDuty

manasmbellani:
Added a new module to print Guardduty findings and stats, Fixed bug in enumeration for Guardduty detector. 

Please note in this case the fix in `detection__enum_services` is necessary to ensure that pacu attempts to pull Guardduty detectors correctly and with appropriate error handling from all regions. Without the fix, the `guardduty__list_findings` will simply not work correctly.
    
    * Added a new module that uses detection__enum_services module to determine all detectors for GuardDuty that are enabled and pulls the statistics and all findings for the specific detector.
    
    * Fixed a bug in pre-req module detection__enum_services which was failing when enumerating AWS fips region for GuardDuty Detector IDs by adding error handling to ensure that  module continues to run and writes data to back-end database


seb1055:
Thanks for the submission, this is a module I would like to get merged soon. The output of the module is very verbose but the output is hard to follow. My recommendation is to save the alerts to a json file in ./downloads/Guardduty and only output the summary for each region. This would make the output more concise. Looking forward to hearing back, thanks!

manasmbellani:
Thanks for the review, @seb1055. I will make another push later today to write the data to the JSON file. Will let you know once done.

manasmbellani:
@seb1055  Updated #193  code with a new commit which writes guard-duty findings to a JSON outfile, instead of to STDIN.

RyanJarv (approved):
Sorry about the delay on getting this merged.

FYI just made a change to the output directory, the json files get written to sessions/<session>/guarduty now.

PR #198, opened 2020-07-30
added a new module for guardduty__list_accounts, for determining the lateral movement scope

manasmbellani:
which includes: 
* Lists the organizational member accounts for GuardDuty
* Lists the Guardduty member accounts
* Lists all the Guardduty member accounts


RyanJarv (approved):
Looks good. Thinking it may make sense to combine this with other methods of getting related account ID's at some point.

Going to merge this with some changes I made some changes to reduce the output from this module. Let me know if any of these changes breaks your use case though.

PR #199, opened 2020-08-12
Added a new module acm__enum to enumerate info in ACM

manasmbellani:
acm__enum/main.py: added a new module to list private CA, certs from ACM which are useful during pentest for determining subdomains belonging to organization, and if we can generate our own certs in private CA


manasmbellani:
Closing as multiple commits had been merged together. See #200

PR #200, opened 2020-08-13
added a new module ACM which is useful during pentest for determining…

manasmbellani:
… subdomains belonging to organization, and if we have any private CA certs as well. Also lists expired certs that can present an opportunity for takeover and operating while hiding presence.

### DIFF
--- a/modules/acm__enum/main.py
+++ b/modules/acm__enum/main.py
@@ -5,13 +5,14 @@ import copy
 import string
 import random
 import json
+import time
 
 module_info = {
     'name': 'acm__enum',
     'author': 'Manas Bellani',
     'category': 'ENUM',
     'one_liner': 'Enumerate Information about the AWS Certificate Manager',
-    'description': 'This module is used to list and get information about ACM certificates, list expired certificates, and get info about private CAs which can generate certs. Expired certificates can provide an opportunity for takeover if domain has expired OR can be abused for client interaction as well',
+    'description': 'This module is used to list and get information about ACM certificates, list expired certificates, and get info about private CAs which can generate certs. Expired certificates can provide an opportunity for takeover if domain has expired OR can be abused for client interaction as well. All certs, cert chains, CAs discovered is written to \"downloads\" folder in relevant \'sessions\' folder',
     'services': ['ACM'],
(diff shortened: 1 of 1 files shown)


DaveYesland:
Nice idea for a module. Looks like the data from the output is not being stored anywhere? Maybe dump this into the `downloads/<session>/acm__enum` directory for the session into files that make sense for each data item.

When running the command with the `--all` flag I get the error below.

```
  File "pacu.py", line 864, in exec_module
    raise ValueError('The {} module\'s summary is too long ({} characters). Reduce it to 10000 characters or fewer.'.format(module.module_info['name'], len(summary)))
<class 'ValueError'>: The acm__enum module's summary is too long (177976 characters). Reduce it to 10000 characters or fewer.
```

Probably should just output the summary of how many of the various items were found and indicate the directory the user can browse to to see the data.

manasmbellani:
Hi @DaveYesland, thanks for reviewing the module. I have added commit to write info to downloads folder and also limited the summary printed. If this looks satisfactory now, will be great if you could merge it with `master`. OTherwise, please let me know if you have any further changes that you would like me to make.

modules/acm__enum/main.py
```diff
@@ -0,0 +1,254 @@
...
+    outfiles = {}
+    outfiles['certs'] = 'sessions/{}/downloads/acm_enum_certs_{}.json'.format(session.name, now)
+    outfiles['certs_info'] = 'sessions/{}/downloads/acm_enum_certs_info_{}.json'.format(session.name, now)
+    outfiles['certs_chain'] = 'sessions/{}/downloads/acm_enum_certs_chain_{}.json'.format(session.name, now)
+    outfiles['cas'] = 'sessions/{}/downloads/acm_enum_cas_{}.json'.format(session.name, now)
+    outfiles['certs_expired'] =  'sessions/{}/downloads/acm_enum_certs_expired_{}.json'.format(session.name, now)
+
+    # Write the relevant output to the output files
+    for info_type, outfile in outfiles.items():
+        if data[info_type]:
```
seb1055, line 226: Added a check to prevent empty output files from being created.

seb1055 (approved):

PR #202, opened 2020-08-25
Fix GZIP decoding problems in ec2__download_userdata

DaveYesland:
This replaces byte comparison which was improperly implemented with try/except blocks to attempt to decode Gzip compressed userdata in the ec2__download_userdata module.


seb1055:
Resolves #149

PR #207, opened 2020-09-01
Module to enumerate Route53 hosted zones and query logging configurations

aaronrea:
* adds Route53 column to db
* adds module to enumerate Route53 Hosted Zones and query logging config if exists

Sample sanitized output:
```
Pacu (t3:imported-collab) > run route53__enum
  Running module route53__enum...                                                                                                                                        
ZoneID: xxxxxxxxxxxx4I  Name: example.com. Private: False
ZoneID: xxxxxxxxxxxxUN  Name: dev.eample.local. Private: False                                                                                                           
QueryLoggingConfigs:
ZoneID:xxxxxxxxxxxx4I :: CloudWatchLogsLogGroupArn: arn:aws:logs:us-east-1:111111111111:log-group:/aws/route53/xxxxxxxxxxxx4I:*                                         
[route53__enum] route53__enum completed.
[route53__enum] MODULE SUMMARY:
Found 2 hosted zones:
{                                                                                                                                                                          
"xxxxxxxxxxxx4I": {
    "Id": "/hostedzone/xxxxxxxxxxx4I",                                                                                                                                      
    "Name": "example.com.",
    "CallerReference": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",                                                                                                               
    "Config": {                                                                                                                                                                
      "PrivateZone": false
    },                                                                                                                                                                       
    "ResourceRecordSetCount": 3,
    "CloudWatchLogsLogGroupArn": "arn:aws:logs:us-east-1:111111111111:log-group:/aws/route53/xxxxxxxxxxxx4I:*",
    "QueryLoggingConfigId": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
  },
  "xxxxxxxxxxxxUN": {                                                                                                                                                        
  "Id": "/hostedzone/xxxxxxxxxxxxUN",                                                                                                                                     
  "Name": "dev.eample.local.",
    "CallerReference": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
    "Config": {
      "PrivateZone": false
    },
    "ResourceRecordSetCount": 5
  }
}
```


core/models.py
```diff
@@ -88,7 +88,8 @@ class PacuSession(Base, ModelUpdateMixin):
         'VPC',
         'WAF',
         'Account',
-        'AccountSpend'
+        'AccountSpend',
```
seb1055, line 91: Adding new fields will break existing pacu sessions. There are other modules pending so we will have to merge them simultaneously to reduce impact.

modules/route53__enum/main.py
```diff
@@ -0,0 +1,111 @@
...
+}
+
+
+parser = argparse.ArgumentParser(add_help=False, description=module_info['description'])
+
+
+def get_hosted_zones(client):
+    try:
+        hosted_zones = client.list_hosted_zones()['HostedZones']
+    except ClientError as error:
```
seb1055, line 46: This error catch should terminate the module exeuction and return main method. Allowing the exeuction to contiune crashes the module on line 51.

seb1055, line 46: The 'f' is missing for string format printing the literal "{error}" rather than the boto3 error.

seb1055 (changes requested):
This is a great module! There are two small changes to make for intial error catching and the error message. Once we get those fixed, I can get this merged. 

Thank you!

modules/route53__enum/main.py
aaronrea: Good call. Would probably have a similar issue around line 68 so I'm going to move this try/except out to main. I'm still looking thru the docs for an answer but is there a recommended way to fail an entire module gracefully? If either of these functions fail, I think it makes sense to fail the module.

aaronrea: addressed in b4c7565

RyanJarv (approved):
Thanks for the PR!

Sorry this took so long to get to, looks good though.

PR #214, opened 2020-11-12
updated ecs__backdoor_task_def based on Sebastians recommendations 

MakoSec:


RyanJarv:
Haven't gotten the chance to review this yet but I don't think we want to add the executable bit to all the files, adding it to just pacu.py or any other's that are meant to be executed directly makes sense though.

Unsure if individual modules are meant to be executed without first invoking pacu.py first? If so adding them to the main.py files would make sense, otherwise just pacu.py.

MakoSec:
> 
> 
> Haven't gotten the chance to review this yet but glancing over it I don't think we want to add the executable bit to all the files, adding it to just pacu.py or any other's that are meant to be executed directly makes sense though.
> 
> Unsure if individual modules are meant to be executed without first invoking pacu.py first? If so adding them to the main.py files would make sense, otherwise just pacu.py.
> 
> Update: Also worth noting the conflicts will be resolved once you set the permissions back to how they where on those files.

I'm not sure what you mean by "executable bit". Are you talking about the stager shell script?

RyanJarv:
Ok no worries, I can fix it when I get a chance.

But just FYI that's what's happening when you see the mode change from 10644 to 10755 on all the files in the diff. Usually you would only do this for files that you want the system to treat as executable.. i.e run directly without a parser. So like running:

python ./pacu.py vs ./pacu.py

Which is actually something we needed to fix, it's just that this isn't needed for the other files.

Screen Shot 2021-02-09 at 3 39 15 PM

MakoSec:
Ahh okay I see what you mean. I think this may be because I cloned it with WSL. I don't recall ever setting anything as executable. Sorry about that!

RyanJarv:
Not sure if you saw it but created a suggested PR to this branch here: https://github.com/MakoSec/pacu/pull/1

RyanJarv:
Closing this in favor of #242 (which is based off this branch)

PR #216, opened 2020-12-15
[WIP] load existing aws_key enumerated permissions when needed

RyanJarv:
<s>
## Issue

If you are piviting to a new role, for example in https://github.com/RhinoSecurityLabs/cloudgoat/blob/master/scenarios/lambda_privesc/README.md iam__privesc__scan doesn't work as expected since the current key (lambdaManager-role) set does not have the info you gathered as the previous user (chris).

## Notes

I'm not expecting this PR to merged as is, instead opening the PR to get some feedback on the what might be the right way to fix this issue.

For this PR I'm just falling back to loading permissions from the exported confirmed_permissions file if it exists for the current role. Mostly wondering if there is a better place to handle something like this.

I'm not too familiar with the code base so I might update this with other thoughts as I look through things.
</s>


RyanJarv:
I think this would be a separate PR but also thinking about adding support for assuming roles. Probably don't want to tie this into that, since ideally this would work regardless of how the role is assumed.

Haven't looked into if this makes sense but another thought is that permissions in the db for the aws_key could be updated as apart of `iam__enum_permissions`.

RyanJarv:
Forgot about this PR and have been using this branch for more general improvements. I think I'm going to hold off on the other stuff for now. I'll close this PR for now and open another one for these newer changes.

PR #220, opened 2021-01-11
General maintenance and improvements

RyanJarv:
* Dependency changes
            * Removing pytz module
    
                    This was just being used to make the datetime object timezone aware,
                    the same thing can be done using:
    
                            .replace(tzinfo=datetime.timezone.utc)
    
                    Which includes the UTC offset in the datetime object.
            * Set lower bounds on boto3/botocore dependencies.
    * Linting and testing
            * Create python-package.yml GitHub action for linting and testing
            * Linting passing with flake8 on ./*.py and ./core
            * Typing checks with mypy passing on ./*.py and ./core
            * Move .flake8 to setup.cfg and add mypy config
            * Removed ignores from flake8 config and changed line length to 160
            * Updated flake8 to resolve errors I was running into with the pinned version
            * Added makefile with flake8, mypy, lint and test targets
            * Add requirements-dev.txt for dev dependencies
            * Initial pytest tests
    * Other
            * Moved static helper funcs out of Main
            * Update sqlalchemy to fix warning when running certain modules.
    
                    SAWarning: The GenericFunction 'array_agg' is already
                    registered and is going to be overriden.
    
            * Make pacu.py the entrypoint for docker.
              Will behave closer to how it does when running it directly this way.


Dockerfile
```diff
@@ -1,13 +1,16 @@
...
 
 # Install Pacu
 WORKDIR /usr/src/pacu/
 COPY ./requirements.txt ./
 RUN pip install -r requirements.txt
 COPY ./ ./
 
-ENTRYPOINT [ "python3" ]
-CMD ["pacu.py"]
+RUN echo 'AWS_EC2_METADATA_DISABLED=true' >> /etc/profile
```
RyanJarv, line 14: https://blog.ryanjarv.sh/2020/10/17/abusing-the-aws-sdk.html

.flake8
```diff
@@ -1,2 +0,0 @@
-[flake8]
-ignore = E126,E127,E128,E266,E501
```
RyanJarv, line 2: Moved flake8 config to setup.cfg as well as removed these ignores.

Makefile
```diff
@@ -0,0 +1,24 @@
...
+	rm -rf out/
+	rm -rf ./**/**/__pycache__/
+
+test:
+	pytest ./tests
+
+graphs:
+	mkdir -p out
+	cd out && pyreverse -S -f ALL -o dot ../pacu.py ../core
+	docker run -it  -v "${PWD}:/dot" -w /dot dotviz  -Tpng out/classes.dot -o out/classes.png
```
RyanJarv, line 24: Was messing around with making a graph of how classes relate, didn't turn out to be terribly interesting though. I'll probably remove this along with the dependency.

setup.cfg
```diff
@@ -0,0 +1,26 @@
+[flake8]
+max-line-length = 160
+exclude = out/*,venv/*,modules_archive/*,.mypy_cache/*,sessions/*
+
+[mypy]
+python_version = 3.9
+follow_imports = skip
+plugins = sqlmypy
+
+[mypy-botocore]
```
RyanJarv, line 10: These are all necessary to avoid mypy from looking at dependencies. Should be a better way to do this but couldn't get that working for some reason.

tests/test_pacu_session.py
```diff
@@ -0,0 +1,50 @@
+import pytest
```
RyanJarv, line 1: Not much here but wanted to at least get an example of how testing might work.

utils.py
```diff
@@ -1,40 +1,41 @@
...
     Session = sessionmaker(bind=engine)
 
     return Session()
 
-
-def remove_empty_from_dict(d):
+def remove_empty_from_dict(d: Union[dict, list, typing.Any]) -> Union[dict, list, typing.Any]:
     """ Reference: https://stackoverflow.com/a/24893252 """
     if type(d) is dict:
+        d = typing.cast(dict, d)
```
RyanJarv, line 27: This is mostly for mypy to know it should be a dict. Don't believe it does anything at runtime.

RyanJarv (commented):
Type annotations make most of the changes here. It's not super common to have these in python yet it seems like but imo they make a huge difference in catching bugs early and understanding what methods you have to work with. They shouldn't be doing anything at runtime but a good editor (pycharm, and I believe vscode as well) should be using them for completion.

MyPy does all the type checking, I added that as a target in the makefile as well as in the github action for linting/testing. Using them isn't required right now though so if you don't want to use them mypy simply treats it as an Any type. 

Using them does add extra imports which can be a problem for circular dependencies, if this comes up it can be avoided by wrapping the annotation in quotes which I believe lazily evaluates it.

RyanJarv:
Reverted removing the awscli dependency, that change doesn't really fit with the rest of this PR.

core/models.py
```diff
@@ -91,7 +92,7 @@ class PacuSession(Base, ModelUpdateMixin):
         'AccountSpend'
     )
 
-    aws_keys = relationship('AWSKey', backref='session', cascade='all, delete-orphan', lazy='dynamic')
+    aws_keys = relationship('AWSKey', back_populates='session', cascade='all, delete-orphan', lazy='dynamic')
```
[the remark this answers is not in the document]
RyanJarv: Tested this change works fine across updates.

seb1055 (approved):
Looks good!

PR #221, opened 2021-01-14
Add iam__pivot integrated

RyanJarv:
This is the integrated version of iam__pivot, it is an attempt to use
 both existing data sources and expose an interface of iam__pivot in
a way that allows us to integrate more escalations.

For a simpler version of this without integration to pacu see
the feature/iam__pivot--basic branch.

One of the idea for this branch is that escalations in iam__privesc
could be included here over time and work alongside the current assume
role escalation.

 i.e you could be presented a path like:

user1 ---           can assume           ---> role2
role2 --- can exploit and gain access to ---> user3
user3 ---           can assume           ---> admin

along with several other's and pick which one you want to traverse.

Checker selection functionality:

  Although this isn't exposed to the user currently adding it
  should be fairly simple with how the integration currently
  works, that is, provided there is more functionality then
  just the sts assume role checker.

  Instead of using seperate checkers we currently have one root class
  which right now is the only one attached to checker_map. We track
  checkers via cls.__subclasses__ which can in turn either be a proxy
  to a larger group of checkers or be a checker it self.

  So attaching a subchecker works the same way as attaching the root,
  and a subchecker can simply use `pass` to act as a group of
  other checkers. In this sense we have a simple way to enable
  escalation checkers in groups, specifically or both.

Performance Improvements:

  Building the graph is already pretty slow, checkers run at O(n),
  while node size is O(n^2) or around there. We'll be adding more
  checker's with this approach, but we can reduce the run time
  by doing a simple precheck's before we make source/dest pairs
  out of the nodes. This is handled through filter_source and
  filter_dest methods.

  In a sense this is a declarative approach to specifying what
  is needed for the escalation to succeed. It may make sense
  to do this where ever possible to simplify in general things
  as well as take advantage of performance improvements like
  mentioned above.

Drawbacks:
  This all sounds great but to get this all to work I had to
  integrate it with pmapper in an unusual way and some of the
  code had to simply be copied to work in a way I needed it
  to.

  I don't think this makes sense, which to be honest is
  more of a gut feeling then anything, to merge this the
  way it is implemented now.

Options:
  To be honest I'm not sure what to do here. After messing
  around with this a bit I feel there's simply more
  important things to focus on in pacu. Maybe over the
  long term it could make sense to integrate some of
  the ideas here into pacu core but at the moment I'm
  not sure what that would look like.

  Anyways this is an interesting proof of concept if
  nothing else.

### DIFF
--- a/modules/iam__enum_users_roles_policies_groups/main.py
+++ b/modules/iam__enum_users_roles_policies_groups/main.py
@@ -36,6 +36,7 @@ parser.add_argument('--users', required=False, action='store_true', help='Enumer
 parser.add_argument('--roles', required=False, action='store_true', help='Enumerate info for roles in the account')
 parser.add_argument('--policies', required=False, action='store_true', help='Enumerate info for policies in the account')
 parser.add_argument('--groups', required=False, action='store_true', help='Enumerate info for groups in the account')
+parser.add_argument('--instance-profiles', required=False, action='store_true', help='Enumerate info for instance profiles in the account')
 
 
 def main(args, pacu_main):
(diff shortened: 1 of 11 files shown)


RyanJarv:
Closing this for now, may come back to it later

PR #222, opened 2021-01-14
Add iam__pivot basic

RyanJarv:
This uses pmmapper to build a graph db of escalation routes and
gives you the option of which one to traverse.

This only checks for assume role but can be extended to check for
other escalations.

This version of the module is fairly basic and doesn't attempt to
integrate with pacu, which means it won't know about previously
found resources and pacu won't know about any resources pmapper
finds.

For a different take on this same module see
features/iam__pivot-integrated.

### DIFF
--- a/modules/iam__pivot/main.py
+++ b/modules/iam__pivot/main.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+import argparse
+import collections
+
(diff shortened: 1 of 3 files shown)


RyanJarv:
Closing this for now, may come back to it later

Issue #233, opened 2021-03-15
ModuleNotFoundError: No module named 'typing_extensions'


Techbrunch:
When trying to run the Docker image I'm running into the following issue:

```
docker run -it rhinosecuritylabs/pacu:latest
Unable to find image 'rhinosecuritylabs/pacu:latest' locally
latest: Pulling from rhinosecuritylabs/pacu
f84cab65f19f: Pull complete
a3d5ab5e63b2: Pull complete
2302001d7ccf: Pull complete
487d5465ac63: Pull complete
bad0f3545de9: Pull complete
50bc5c8cb4f7: Pull complete
fa9aa0d7dfd8: Pull complete
753428fae5a9: Pull complete
bd5ee32eac55: Pull complete
6e7a1e3f55f1: Pull complete
191f36307f25: Pull complete
Digest: sha256:949e03fe67985c84197e997064b222ce0dd8c9634e30fbd548cf81756b4e95f0
Status: Downloaded newer image for rhinosecuritylabs/pacu:latest

settings.py file not found. Creating one from settings_template.py
  Settings file created.

Traceback (most recent call last):
  File "/usr/src/pacu/pacu.py", line 1732, in <module>
    Main().run()
  File "/usr/src/pacu/pacu.py", line 175, in __init__
    self.CATEGORIES: set = load_categories()
  File "/usr/src/pacu/pacu.py", line 61, in load_categories
    module = __import__(module_path, globals(), locals(), ['module_info'], 0)
  File "/usr/src/pacu/modules/ebs__download_snapshots/main.py", line 8, in <module>
    from typing_extensions import TypedDict
ModuleNotFoundError: No module named 'typing_extensions'
```

PR #235, opened 2021-03-15
Adding typing-extensions dep

seb1055:
typing-extensions is not listed in requirements.txt causing Pacu to crash after installing. Added the dep to be installed on setup. 

Fixes issue #233


RyanJarv:
Awesome thanks, lgtm.

As a follow up we may want to look into why the tests didn't catch this, feel like they should have.

RyanJarv (approved):

PR #236, opened 2021-03-24
Module to download CloudFormation data

DaveYesland:
This module downloads all templates, parameters and outputs from CloudFromation. It parses each as they are downloaded for secrets and outputs any findings.

### DIFF
--- a/modules/cloudformation__download_data/main.py
+++ b/modules/cloudformation__download_data/main.py
@@ -1,0 +1,84 @@
+import argparse
+from botocore.exceptions import ClientError
+import json, os, datetime
+from json import JSONEncoder
+from core.secretfinder.utils import regex_checker, Color
+module_info = {'name':'cloudformation__download_data', 
+ 'author':'David Yesland', 
+ 'category':'ENUM', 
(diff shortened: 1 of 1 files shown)


modules/cloudformation__download_data/main.py
```diff
@@ -0,0 +1,86 @@
...
+            if isinstance(obj, (datetime.date, datetime.datetime)):
+                return obj.isoformat()
+
+    stacks = []
+    found_regions = []
+    for region in regions:
+        client = pacu_main.get_boto3_client('cloudformation', region)
+        print('Looking for CloudFormation Stacks in region {}...'.format(region))
+        stacks_data = client.describe_stacks()
+        stacks += stacks_data['Stacks']
```
RyanJarv, line 46: Looks like we end up getting duplicate reports in the output because this isn't reset for each region.

The for loop below ends up getting executed for each stack in all the regions run so far.

RyanJarv (changes requested):

RyanJarv:
Made this PR to your branch that fixes the duplicate downloads.

https://github.com/DaveYesland/pacu/pull/1

RyanJarv (approved):

PR #237, opened 2021-04-03
API Gateway Enumeration Module 

seb1055:
The module starts by looping over all available APIs. For each API it will enumerate the available routes, methods, authentication, and associated services like Lambda. The module will then attempt to enumerate available API keys and client certificates. Finally, the module will attempt to export the definitions as swagger files. 

All results including the swagger exports will be saved to the session's download file. 

The module does not support HTTP or WebsocketAPI types. These types use the boto3 library "apigatewayv2" and use and different methods and response structures. I currently don't plan on adding support for these types.

### DIFF
--- a/modules/enum__apigateway/main.py
+++ b/modules/enum__apigateway/main.py
@@ -2,22 +2,21 @@
 
 import argparse
 import os
-import sys
 from pathlib import Path
-from pacu import Main
 import pprint
 import json
-import time 
+import time
 
 module_info = {
     'name': 'enum__apigateway',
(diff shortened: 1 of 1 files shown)


RyanJarv (approved):
Awesome, thanks for the PR!

Looks good, just made a few minor changes around linting and how the summary report is generated.

RyanJarv:
@seb1055 Just FYI I updated the module name to be apigateway_enum to match the other enum modules.

seb1055:
Awesome thanks for the quick turnaround and updates!

PR #239, opened 2021-04-05
Fix ecs__enum_task_def datetime serialization

RyanJarv:
The datetime object in the ecs task def response causes serialization to fail without setting a default method.

This change sets the default to str.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -61,7 +61,7 @@ def stringify(obj: Union[dict, list, datetime]) -> Union[dict, list, str]:
         return str(obj.strftime("%a, %d %b %Y %H:%M:%S"))
 
     elif isinstance(obj, bytes):
-        # If obj is a datetime, return a formatted string version of it
+        # If obj is bytes, return a formatted string version of it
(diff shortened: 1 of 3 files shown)


utils.py
```diff
@@ -35,31 +35,35 @@ def remove_empty_from_dict(d: Union[dict, list, typing.Any]) -> Union[dict, list
...
-            new_list.append(stringify_datetime(v))
+            new_list.append(stringify(v))
         return new_list
 
     elif isinstance(obj, datetime):
         # If obj is a datetime, return a formatted string version of it
         return str(obj.strftime("%a, %d %b %Y %H:%M:%S"))
 
+    elif isinstance(obj, bytes):
+        # If obj is a datetime, return a formatted string version of it
```
DaveYesland, line 64: I think this should be:
`# If obj is bytes, return a formatted string version of it`

DaveYesland (changes requested):

DaveYesland (approved):

PR #244, opened 2021-04-15
Dev Dockerfile and iam__enum_roles improvements

### LINKED ISSUE #229: iam__enum_roles - Role trust isn't restored - brakes role
Hi!

First of all I want to thank you for this great project.
I have encountered an interesting problem when I run the module: iam__enum_roles in our AWS environment.
The current script, as far as I understand it, exploits AWS verification error if a statement with an invalid principle is entered - hence flagging the error: MalformedPolicyDocumentException.
While if a role enumeration is successful, and such a role exists, the policy update of the trust relationships succeeds.

What we noticed however is that the original trust relationship doesn't seem to be stored in the script you use, hence leaving the role used for enumeration with an invalid trust relationship after the script finishes.
This is especially bad if the role entered is a federated role, which can then brake the access for an entire group that uses that role (which is what happened in our case, although for a sandbox account).

Another thing we noticed is that if the script doesn't finishes in time, if the token expires before completion, this will also brake the role.

I humbly suggest the following changes:
1) Add a warning text when entering the --role-name parameter to verify that it's not a federated role and that the aws secret key added in the key database must have at least 1 hour until expiration when the script is started.

2) Store the original trust relationship policy in a parameter so that it can be restored once all enumeration tests are done.
This can be done by adding the following:

client = boto3.client('iam')
response = client.get_role(
    RoleName=role_name
)
AssumeRolePolicyDocument=response["Role"]["AssumeRolePolicyDocument"]

RyanJarv:
* Add's Dockerfile.dev which also installs packages in requirements-dev.txt
* iam__enum_roles improvements
  * Rolls back changes on updated role at the end of the run
  * If --role-arn isn't passed use a temporary role instead
  * Fixes #229

Would be good to start adding tests to some of these, may have to come back to that though.

### DIFF
--- a/modules/iam__enum_roles/main.py
+++ b/modules/iam__enum_roles/main.py
@@ -1,39 +1,58 @@
 #!/usr/bin/env python3
 import argparse
+import sys
+
 import botocore
+import botocore.exceptions
 import random
 import string
 
+import typing
+
+import json
+
+if typing.TYPE_CHECKING:
+    import mypy_boto3_iam
+    import mypy_boto3_iam.type_defs
 
 module_info = {
     'name': 'iam__enum_roles',
-
     'author': 'Spencer Gietzen of Rhino Security Labs',
-
     'category': 'RECON_UNAUTH',
-
     'one_liner': 'Enumerates IAM roles in a separate AWS account, given the account ID.',
-
-    'description': 'This module takes in a valid AWS account ID and tries to enumerate existing IAM roles within that account. It does so by trying to update the AssumeRole policy document of the role that you pass into --role-name. For your safety, it updates the policy with an explicit deny against the AWS account/IAM role, so that no security holes are opened in your account during enumeration. NOTE: It is recommended to use personal AWS access keys for this script, as it will spam CloudTrail with "iam:UpdateAssumeRolePolicy" logs and a few "sts:AssumeRole" logs. The target account will not see anything in their logs though, unless you find a misconfigured role that allows you to assume it. The keys used must have the iam:UpdateAssumeRolePolicy permission on the role that you pass into --role-name to be able to identify a valid IAM role and the sts:AssumeRole permission to try and request credentials for any enumerated roles.',
-
+    'description': (
+        'This module takes in a valid AWS account ID and tries to enumerate existing IAM roles within that account. It does '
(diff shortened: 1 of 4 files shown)


modules/iam__enum_roles/main.py
```diff
@@ -48,32 +67,44 @@ def main(args, pacu_main):
...
+                RoleName=role_name,
+                PolicyDocument='''
+                {{
+                    "Version":"2012-10-17",
+                    "Statement":[{{
+                        "Effect":"Deny",
+                        "Principal":{{"AWS":"{}"}},
+                        "Action":"sts:AssumeRole"
+                    }}]
+                }}'''.format(role_arn).strip()
```
RyanJarv, line 101: Double brackets are needed here to escape interpolation with the format function.

modules/iam__enum_roles/main.py
```diff
@@ -124,7 +153,48 @@ def main(args, pacu_main):
...
+                    "Effect": "Deny",
+                    "Principal": {"AWS": "*"},
+                    "Action": "sts:AssumeRole"
+                }]
+            }
+            '''.strip(),
+        )
+
+    try:
+        run(args, role_name, pacu_main, iam)
```
RyanJarv, line 183: Code from main was moved to run so we can cleanup the role if it fails for whatever reason.

modules/iam__enum_roles/main.py
```diff
@@ -124,7 +153,48 @@ def main(args, pacu_main):
...
+            print(f"Reverting the {role_name} trust policy.")
+            try:
+                iam.update_assume_role_policy(RoleName=role_name, PolicyDocument=orig_trust_doc)
+            except Exception as err:
+                print(f"unable to revert {role_name} to it's original state. this can be done manually by updating the assume"
+                      f"role document to the following: \n\n{orig_trust_doc}\n\n")
+                print(err)
+        else:
+            print(f"Cleaning up the {role_name} role.")
+            iam.delete_role(RoleName=role_name)
```
RyanJarv, line 197: Revert the trust policy doc if args.role_name was passed otherwise delete it since it's a temporary role created for this run.

DaveYesland (approved):

Issue #230, opened 2021-02-10
installation as a package ?


sbrun:
Hello,

I started to work on packaging pacu for Kali linux (https://bugs.kali.org/view.php?id=7030)

From what I see you don't provide a way to install in a system wide manner (FHS compliant).
Several files/directories are created or rewritten in pacu directory (like settings.py or global_error_log.txt files) and in sessions and modules directories.
If I installed pacu in /usr/share/pacu (like it should be I think), the user will have no access to modify the files and he will not be able to use pacu.
Do you have recommendations to install pacu in this way?

If you need to write files they should in $HOME/.local/share or maybe other directories depending on the use of the files:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

RyanJarv:
I'll have to look into if there's a straightforward way of doing this.

Worst case if this isn't we are looking into potentially refactoring Pacu, If that happens first this would be a good thing to include.

RyanJarv:
This is still a work in progress but linked the PR I'm using for moving to a common output/config directory.

I think for now I'd like to avoid moving things around to much so I'm essentially just adding moving things to ~/.local/share/pacu through a common setting. I'm not sure how much it makes sense to strictly adhere to FHS beyond that (or is that enough?), but I think either way we can come back and change things later. Most of the changes here is just un-hardcoding everything.

@sbrun Let me know if you have any thoughts on this.

RyanJarv:
One other thing I'm wondering, just in case you see this before I dig into it much is how Kali Linux packages pip packages. If this was installable through pip would that be enough to get it set up through dpkg or is there any thing else to be aware of there?

RyanJarv:
@sbrun I'm testing packaging with dh-python on https://github.com/RhinoSecurityLabs/pacu/tree/test/debian-packaging (debian files there where copied from another Kali python package and likely are not completely correct).

This seems to more or less work (needs more testing still) but one issue I could use some input on is we rely on https://github.com/RhinoSecurityLabs/dsnap for one of the modules which seems to get omitted when building the package, I imagine this is because this only exists in pip.

Does that need to be packaged as well and if so will the pyproject.toml style setup work here (as opposed to using setuptools)?

I'll likely be merging #247 in sometime soon (after more testing) which should take care of the output directory issues. This also get's things setup so we can push Pacu to PyPi.

If you can let me know if the dsnap dependency will be an issue and if there is anything I can do there to help that would be appreciated.

[...]

PR #247, opened 2021-04-17
Support for packaging

RyanJarv:
To support packaging and installing along other packages in a sane way this commit moves all the config and output files to `~/.local/share/pacu`. Pacu python source files are also moved under a pacu subdirectory (needed for setuptools).

This is working but needs more testing. When this is done it will Resolve: #230.

Some things where annoying me a bit so did some other cleanup while working on this, mostly shortening long lines and things like that.


core/lib.py
```diff
@@ -0,0 +1,58 @@
+import contextlib
+import io
+import os
+from typing import Optional, cast
+
+from pathlib import Path
+
+import settings
+
+get_active_session: Optional[callable] = None
```
RyanJarv, line 10: This is a hack so we can get the session name without needing to pass around pacu_main everywhere.

modules/cloudtrail__download_event_history/main.py
```diff
@@ -3,30 +3,23 @@
...
-    'description': 'This module will download the CloudTrail event history for each specified region in both JSON format to ./sessions/[current_session_name]/downloads/cloudtrail_[region]_event_history_[timestamp].json. Warning: This module can take a very long time to complete because the maximum events per API call is 50, when there could be tens or hundreds of thousands or more total events to download. A rough estimate is about 10000 events retrieved per five minutes.',
-
-    # A list of AWS services that the module utilizes during its execution
+    'description': strip_lines('''
+        This module will download the CloudTrail event history for each specified region in both JSON format to 
+        ~/.local/share/pacu/sessions/[current_session_name]/downloads/cloudtrail_[region]_event_history_[timestamp].json.
+        Warning: This module can take a very long time to complete because the maximum events per API call is 50, when 
+        there could be tens or hundreds of thousands or more total events to download. A rough estimate is about 10000
+        events retrieved per five minutes.
+    '''),
```
RyanJarv, line 20: This doesn't show up in GitHub too well but before this was just a single long line. I suppose soft wrapping was probably enabled before, but with python's explicit white space it's best to make this explicit.

RyanJarv:
Assuming we don't run into any issues here I think this is good now. Pushed a package to PyPi we can use for testing for now (before it was just a placeholder).

```
pip install -U pip
pip install -U pacu # Should install 0.1.12
pacu
```

RyanJarv:
Currently lambda__backdoor_new_users will fail with:
```
    <class 'FileNotFoundError'>: [Errno 2] No such file or directory: '/home/test/.local/share/pacu/test/module/lambda__backdoor_new_users/lambda_function.py'
```

pacu/modules/systemsmanager__download_parameters/main.py
```diff
@@ -1,126 +1,126 @@
...
+
+    print = pacu_main.print
+
+    args = parser.parse_args(args)
+
+    if args.regions:
+        regions = args.regions.split(',')
+    else:
+        get_regions = pacu_main.get_regions
+        regions = get_regions('ssm')
```
RyanJarv, line 39: Need to look into what's going on here, probably line endings changed.

RyanJarv:
This is ready to merge. There's a fair amount of style related changes, I tried moving those out to a different PR but unfortunately not as simple as running `git reset -p` and reverting specific changes. I can look into this more if it's an issue though, or if this is just too much to review (@DaveYesland).

I'll go through and add comments to things shortly, but a quick overview of other changes:
* Uses poetry for building/pushing to pip (dev only dependency).
  * requirements* still exist, but generally won't be needed (but exist to support installing the old way) and are now generated from pyproject.toml/poetry.lock.
* Add's Dockerfile.dev for running dev related tasks
* Add's GitHub actions to deploy to PyPi.
  * This can be triggered locally by running ./scripts/publish.sh which just run's some sanity checks and push's a tag to master.

configure_settings.py
```diff
@@ -1,17 +0,0 @@
-#!/usr/bin/env python
```
RyanJarv, line 1: This file doesn't exist anymore. Don't believe there's really any need for dynamically generating settings, it's always the same anyways.

cli.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
```
RyanJarv, line 1: Exists to support the old way of running pacu. Unfortunately we can't name this pacu.py because various tools get confused when this is named the same as the module.

I believe people will figure this out, or worst case look at the README.md (which then suggests installing through pip) when it doesn't exist anymore. But, just thinking now, maybe we could maybe add a note to the top of the README.md as well.

last_update.txt
```diff
@@ -1 +0,0 @@
-2020-05-08
```
RyanJarv, line 1: Moved to pacu/last_update.txt

requirements.txt
```diff
@@ -1,10 +1,134 @@
-awscli
-dsnap>=0.1.10
-boto3>=1.15
-botocore>=1.16
-s3transfer>=0.3.0
-requests==2.25.1
-urllib3==1.26.4
-SQLAlchemy==1.3.22
-SQLAlchemy-Utils==0.36.8
-typing-extensions==3.7.4.3
```
RyanJarv, line 10: This lives in pyproject.toml now, this file is generated from that with `poetry export`

requirements-dev.txt
```diff
@@ -1,8 +1,245 @@
-boto3>=1.15
```
RyanJarv, line 1: Same as requirements.txt, moved to pyproject.toml. This is not generated from that with `poetry export --dev`.

pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[tool.poetry]
+name = "pacu"
+version = "0.1.15"
+description = ""
+authors = ["Ryan Gerstenkorn <ryan.gerstenkorn@rhinosecuritylabs.com>"]
```
RyanJarv, line 5: Not really sure what makes sense to be here.. I think this is just what shows up on pip. Might make sense to put pacu@rhinosecuritylabs.com instead.

DaveYesland (approved):

DaveYesland (approved):

PR #248, opened 2021-04-29
feat: rds enumeration module

juliosmelo:
This module enumerates all relevant instances databases from AWS RDS of a given region, including databases master username, SGDB (engine), port and endpoints


RyanJarv:
Thanks for the PR!

This looks good. Just one minor thing, can we change the directory name to rds__enum (with two underscores)? Realize it's a bit strange, but this will keep things consistent with the other modules.

Also worth noting if you get an error about regions that's fixed by this PR https://github.com/RhinoSecurityLabs/pacu/pull/250.

RyanJarv (changes requested):
Mentioned above but just need to rename the directory to rds__enum with two underscores.

juliosmelo:
Done! Thank you mate!

RyanJarv:
Thanks for getting this updated with the new structure! Realize that switched things up a bit here.

Since there's another RDS module in the dev branch I'm going to go ahead and merge this there then we can merge this in to master along side that.

RyanJarv (approved):
lgtm

PR #252, opened 2021-05-09
Eks enum

bomjumaku:
A module to enumerate EKS resources.


pacu/modules/eks__enum/main.py
```diff
@@ -0,0 +1,85 @@
...
+        print(f"clusters in {region}: {clusters}")
+        if len(clusters) > 0:
+            cluster_count += len(clusters)
+            data[region] = {
+            "clusters": {}
+            }
+            for cluster in clusters:
+                if args.verbose:
+                    data[region]['clusters'][cluster] = {
+                        "cluster_description": eks_client.describe_cluster(name=cluster)["cluster"],
```
RyanJarv, line 61: Any reason this isn't included by default? I think I'd recommend doing that and removing the verbose flag.

pacu/modules/eks__enum/main.py
```diff
@@ -0,0 +1,85 @@
...
+                else:
+                    data[region]['clusters'][cluster] = {
+                        "nodegroups": eks_client.list_nodegroups(clusterName=cluster)["nodegroups"]
+                    }
+                    if args.addons or args.all:
+                        data[region]['clusters'][cluster]["addons"] = eks_client.list_addons(clusterName=cluster)["addons"]
+                    if args.fargate_profiles or args.all:
+                        data[region]['clusters'][cluster]["fargate_profiles"] = eks_client.list_fargate_profiles(clusterName=cluster)["fargateProfileNames"]
+                    if args.identity_provider_configs or args.all:
+                        data[region]['clusters'][cluster]["identity_provider_configs"] = eks_client.list_identity_provider_configs(clusterName=cluster)["identityProviderConfigs"]
```
RyanJarv, line 79: These all need pagination, otherwise we'll only see partial results in some cases. Looks like the ECS resource has paginators for this, so that may help here (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#paginators).

pacu/modules/eks__enum/main.py
```diff
@@ -0,0 +1,85 @@
...
+    else:
+        regions = args.regions.split(',')
+    for region in regions:
+        eks_client = pacu_main.get_boto3_client('eks', region)
+        clusters = eks_client.list_clusters()["clusters"]
+        print(f"clusters in {region}: {clusters}")
+        if len(clusters) > 0:
+            cluster_count += len(clusters)
+            data[region] = {
+            "clusters": {}
```
RyanJarv, line 56: This line should be indented.
```suggestion
                "clusters": {}
```

pacu/modules/eks__enum/main.py
```diff
@@ -0,0 +1,85 @@
...
+}
+
+
+parser = argparse.ArgumentParser(add_help=False, description=module_info['description'])
+parser.add_argument('--regions', required=False, default=None, help='One or more (comma separated) AWS regions in the format "us-east-1". Defaults to all session regions.')
+parser.add_argument('--verbose', required=False, action='store_true', default=False, help='Enable verbose output')
+parser.add_argument('--addons', required=False, action='store_true', default=False, help='Enumerate EKS addons')
+parser.add_argument('--identity_provider_configs', required=False, action='store_true', default=False, help='Enumerate EKS identity provider configs')
+parser.add_argument('--fargate_profiles', required=False, action='store_true', default=False, help='Enumerate EKS fargate profiles')
+parser.add_argument('--all', required=False, action='store_true', default=False, help='Enumerate all EKS resources.')
```
RyanJarv, line 32: I think I would default to all resources and have the other, more specific flags override this.

RyanJarv (changes requested):
Hi, thanks for the PR!

Added a few comments on some things that need to be changed. I think mainly we need to make sure pagination is supported otherwise we might be returning partial results, the rest are more minor suggestions.

We may want to come back and add describe call's to this module but I don't think that's strictly necessarily for this PR if it's not needed right now. Really the enum modules may be getting a rework sometime soon to make this simpler so I wouldn't worry too much about it right now.

RyanJarv:
Just glancing at the older commits, I like the idea of enumerating the EKS cluster it self. We hope to add support for other providers other than AWS, so think we'll have to keep this use case in mind when we get a chance to dig into that more.

settings.py
```diff
@@ -0,0 +1,21 @@
+import os
```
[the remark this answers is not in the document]
bomjumaku: Good catch.

pacu/modules/eks__enum/main.py
bomjumaku: Yeah, I've not used pacu a lot at this point, so I'm kind of just imagining use cases while writing this. The verbose behavior is now default.

pacu/modules/eks__enum/main.py
bomjumaku: Suggestion implemented. :+1:

pacu/modules/service_regions.json
```diff
@@ -1 +1 @@
...
...n": "us-east-2"}, "hostname": "waf-regional.us-east-2.amazonaws.com"}, "us-west-1": {"credentialScope": {"region": "us-west-1"}, "hostname": "waf-regional.us-west-1.amazonaws.com"}, "us-west-2": {"credentialScope": {"region": "us-west-2"}, "hostname": "waf-regional.us-west-2.amazonaws.com"}}}, "workdocs": {"endpoints": {"ap-northeast-1": {}, "ap-southeast-1": {}, "ap-southeast-2": {}, "eu-west-1": {}, "us-east-1": {}, "us-west-2": {}}}, "workmail": {"endpoints": {"eu-west-1": {}, "us-east-1": {}, "us-west-2": {}}}, "workspaces": {"endpoints": {"ap-northeast-1": {}, "ap-northeast-2": {}, "ap-south-1": {}, "ap-southeast-1": {}, "ap-southeast-2": {}, "ca-central-1": {}, "eu-central-1": {}, "eu-west-1": {}, "eu-west-2": {}, "sa-east-1": {}, "us-east-1": {}, "us-west-2": {}}}, "xray": {"endpoints": {"af-south-1": {}, "ap-east-1": {}, "ap-northeast-1": {}, "ap-northeast-2": {}, "ap-northeast-3": {}, "ap-south-1": {}, "ap-southeast-1": {}, "ap-southeast-2": {}, "ca-central-1": {}, "eu-central-1": {}, "eu-north-1": {}, "eu-south-1": {}, "eu-west-1": {}, "eu-west-2": {}, "eu-west-3": {}, "me-south-1": {}, "sa-east-1": {}, "us-east-1": {}, "us-east-2": {}, "us-west-1": {}, "us-west-2": {}}}}
```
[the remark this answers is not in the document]
bomjumaku: I'm still not sure what this change was. I couldn't see any difference when looking over it visually, but I moved in the version of the file from master which seems to have resolved this.

pacu/modules/eks__enum/main.py
bomjumaku: Sounds good. Working on pagination now.

bomjumaku: Pagination should be implemented with this last commit. Running some rudimentary testing now.

RyanJarv (approved):
Awesome, looks good. Since this makes a change to the sqlitedb schema I'm going to merge it into dev for the moment.

bomjumaku:
Roger that, let me know if there are any issues caused by that schema change. It only caused problems for me if I changed the schema after the initial run. 

ie, I had to delete the database after every schema change.

PR #254, opened 2021-05-14
Fix save()

berney:
Fix the use of `save()` after the packing refactor.

### DIFF
--- a/pacu/modules/cloudwatch__download_logs/main.py
+++ b/pacu/modules/cloudwatch__download_logs/main.py
@@ -6,7 +6,7 @@ import os
 
 from botocore.exceptions import ClientError
 
-from pacu.core.lib import strip_lines, downloads_dir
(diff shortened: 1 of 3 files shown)


RyanJarv (approved):
Thanks for the PR!

Looks like there is still an issue with iam__privesc_scan --offline but we can look into that separately.

RyanJarv:
Python 3.9 failure looks unrelated to this.

berney:
> Thanks for the PR!
> 
> Looks like there is still an issue with iam__privesc_scan --offline but we can look into that separately.

You're welcome. I really like Pacu, so thanks for developing it.

What issue did you see with `iam__privesc_scan --offline`?
I made this PR to resolve the `save` not found issue I hit, and it seemed to work fine for me, producing a json file. It would go in `downloads/downloads` which is probably a small bug.

Issue #226, opened 2021-01-26
iam__enum_permissions can't determine current user


RyanJarv:
# Problem

When running iam__enum_permissions, the module often can't determine the current user and requires you to put it manually. Even when the current user is already known.

# Example

```
Pacu (test:imported-test-user) > whoami
{
  "UserName": "test-user",
  "RoleName": null,
...
```

```
Pacu (test:import-test-user) > run iam__pivot
  Running module iam__pivot...
[pivot]   Running module iam__enum_permissions...
[iam__enum_permissions] Failed to discover the current users username, enter it now or Ctrl+C to exit the module
```

# Fixes

Unsure why Pacu thinks the user isn't currently known, but if we can get that info from the same source as the whoami output it likely makes sense to do that.

Besides that there's many ways to determine the current user via the API, think we could handle this better as well as make the functionality available via an imported function or PacuMain.

RyanJarv:
This should be resolved with the PR above, sts get-caller-identity is generally available even when other permissions aren't.

PR #255, opened 2021-05-16
fix iam__enum_permissions can't determine current user (#226)

naikordian:
iam__enum_permissions used `GetUser` API to get a username, but sometime the user doesn't have a `GetUser` permission
```python
            client = pacu_main.get_boto3_client('iam')
            try:
                user = client.get_user()
                active_aws_key.update(
                    pacu_main.database,
                    user_name=user['User']['UserName'],
                    arn=identity['Arn'],
                    user_id=identity['UserId'],
                    account_id=identity['Account']
                )
````
This pull request will extract username from user's ARN

### DIFF
--- a/pacu/modules/iam__enum_permissions/main.py
+++ b/pacu/modules/iam__enum_permissions/main.py
@@ -130,7 +130,7 @@ def main(args, pacu_main: 'Main'):
             username = identity['Arn'].split(':user/')[1]
             active_aws_key.update(
                 pacu_main.database,
-                user_name=username,
+                user_name=username.split('/')[-1],
                 arn=identity['Arn'],
                 user_id=identity['UserId'],
                 account_id=identity['Account']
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by RyanJarv** [comment]
@naikordian Thanks for the PR!

Switching to use get_caller_identity makes sense to me. The one thing I'd like to add to this is handling user/role paths (so the ARN may have multiple '/'s in the name and we just want the last part here).

File: pacu/modules/iam__enum_permissions/main.py [hunk 1]
```diff
@@ -125,35 +125,16 @@ def main(args, pacu_main: 'Main'):
...
-                        user_id=identity['UserId'],
-                        account_id=identity['Account']
-                    )
-                    return False
+            # GetCallerIdentity away return user's ARN like this if it was a user
+            # arn:aws:iam::123456789012:user/username
+            username = identity['Arn'].split(':user/')[1]
+            active_aws_key.update(
+                pacu_main.database,
+                user_name=username,
```
RyanJarv, line 133: We'll need to verify this works but thinking something like this:

```suggestion
                user_name=username.split('/')[-1],
```

RyanJarv (changes requested):

pacu/modules/iam__enum_permissions/main.py
RyanJarv: Looks good

RyanJarv (approved):
lgtm

PR #256, opened 2021-05-18
fix #218 iam__enum_users is not adding users to database

### LINKED ISSUE #218: Module iam__enum_users is not adding users to database
After running the module `iam__enum_users` successfully, and finding at least one user, the user(s) aren't being added to the database. `data IAM` returns the below.

```
 > data IAM
{
  "Users": []
}

```

naikordian:
iam__enum_roles has a similar bug as well, I will it fix later

### DIFF
--- a/pacu/modules/iam__enum_users/main.py
+++ b/pacu/modules/iam__enum_users/main.py
@@ -32,8 +32,6 @@ parser.add_argument('--account-id', required=True, help='The AWS account ID of t
 
 
 def main(args, pacu_main):
-    session = pacu_main.get_active_session()
-
     args = parser.parse_args(args)
     print = pacu_main.print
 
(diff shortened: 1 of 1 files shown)


RyanJarv:
Thanks for the PR! We should be able to take a look at this sometime this week.

RyanJarv (changes requested):
Hey, looks good. I ran into a few issues when testing this, however for the most part that's not related to this PR and I can fix that elsewhere.

One thing I did notice though is the user structure doesn't match what's used in other modules. Made a suggestion that fixes that, if that looks good to you we can get this merged.

naikordian:
I will fix  #257  later if this PR passes review.

RyanJarv (approved):
lgtm

PR #257, opened 2021-05-19
fix iam__enum_roles is not adding roles to the database

naikordian:


pacu/modules/iam__enum_roles/main.py
```diff
@@ -154,6 +155,19 @@ def run(args, role_name, pacu_main, iam):
...
+def update_roles_database(pacu_main, roles):
+    session = pacu_main.get_active_session()
+    iam_data = deepcopy(session.IAM)
+    if iam_data.get('Roles') is None:
+        iam_data['Roles'] = roles
+    else:
+        # Check for duplicate roels
+        for role in roles:
+            if role not in iam_data['Roles']:
+                iam_data['Roles'].append(role)
```
RyanJarv, line 167: Same thing as in the users module. Haven't tested this but believe it should work based on what I tried for the other module.

```suggestion
                if role not in [r['Arn'] for r in iam_data['Roles']]:
                    iam_data['Roles'].append({
                        "Arn": role,
                        "AssumeRolePolicyDocument": None,
                        "CreateDate": None,
                        "Description": None,
                        "MaxSessionDuration": None,
                        "Path": "/" + '/'.join(role.split(':')[-1].split('/')[1:-1]),
                        "RoleId": None,
                        "RoleName": role.split('/')[-1]
                    }
```

RyanJarv (changes requested):
Thanks for the PR! Looks good, just noticed one thing which is mostly the same as in the other user module PR (https://github.com/RhinoSecurityLabs/pacu/pull/256#pullrequestreview-674746431).

RyanJarv (approved):
lgtm. Made a change to fix the default word file path, but looks like that was already fixed in another PR so ended up reverting it.

PR #260, opened 2021-05-26
Added 'data' command accepts case insensitive service name

naikordian:
It's more convenient if 'data' command accepts both upper case and lower case services name


RyanJarv (approved):
Thanks for the PR! Agree this is a needed change.

PR #264, opened 2021-05-31
EKS enumeration module

juliosmelo:


RyanJarv:
Appreciate the PR! We unfortunately just got a EKS module merged into the dev branch however so we may not be able to merge this as is. I'll try to take a look into this soon to see if we can reuse anything you have here for that module though.

juliosmelo:
Ops... no problem. I did not check the dev branch before sending the PR. Thanks for your feedback.

RyanJarv:
I think we could improve visibility on what's being worked on as well, I'll see what we can do there.

FYI I'm keeping this open to see if we can use anything from this version in that module, just have been a bit busy lately so haven't gotten back around to this yet.

PR #268, opened 2021-06-02
Fixes for iam__enum users/roles modules

RyanJarv:
* Update the path of the default-word-list.txt file for the new
  pacu structure.
* Other minor improvements.

### DIFF
--- a/pacu/modules/iam__enum_roles/main.py
+++ b/pacu/modules/iam__enum_roles/main.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import argparse
 import sys
+from pathlib import Path
 
 import botocore
 import botocore.exceptions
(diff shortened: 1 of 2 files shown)


pacu/modules/iam__enum_roles/main.py
```diff
@@ -111,7 +112,7 @@ def run(args, role_name, pacu_main, iam):
                 return data
             else:
                 print('  Unhandled error: {}'.format(str(error)))
-                return data
+                raise error
```
RyanJarv, line 115: I don't think there's any reason to return the data here if the module fails, doing so also prevents the error from being logged.

pacu/modules/iam__enum_roles/main.py
```diff
@@ -159,7 +160,7 @@ def main(args, pacu_main):
     iam: 'mypy_boto3_iam.IAMClient' = pacu_main.get_boto3_client('iam')
 
     if args.role_name:
-        role_name = args.role_name
+        role_name = args.role_name.split('/')[-1]  # Handle ARN's if that was passed for whatever reason.
```
RyanJarv, line 163: Works the same as before but handles cases where a role ARN was passed instead of a name.

DaveYesland (approved):

PR #270, opened 2021-06-05
Add a region argument for module 'cloudwatch__download_logs'

naikordian:
I don't wanna downloads Cloudwatch logs in all regions. That a lot of logs


RyanJarv (approved):
lgtm

PR #271, opened 2021-06-12
Fix bug data command not found some service and improve minor UX.

naikordian:
https://github.com/RhinoSecurityLabs/pacu/blob/21cd33759870a5f84a2380cd94ab1ec2dcaed823/pacu/main.py#L688-L694

#260 has a bug when some service name didn't use all upper case latter
```
APIGateway      CloudTrail      CloudWatch      CodeBuild
```

### DIFF
--- a/pacu/main.py
+++ b/pacu/main.py
@@ -685,9 +685,13 @@ class Main:
             self.print('\nSession data:')
             session.print_all_data_in_session()
         else:
-            service = command[1].upper()
+            service = command[1]
+            if service.upper() in session.aws_data_field_names:
+                service = service.upper()
+
             if service not in session.aws_data_field_names:
-                print('  Service not found.')
+                print('  Service not found. Please use the service name below.')
+                print('\t'.join(list(session.aws_data_field_names)))
             elif getattr(session, service) == {} or getattr(session, service) == [] or getattr(session, service) == '':
                 print('  No data found.')
             else:
--- a/pacu/main.py
+++ b/pacu/main.py
@@ -685,17 +685,15 @@ class Main:
             self.print('\nSession data:')
             session.print_all_data_in_session()
         else:
-            service = command[1]
-            if service.upper() in session.aws_data_field_names:
-                service = service.upper()
-
-            if service not in session.aws_data_field_names:
+            service = command[1].upper()
+            service_map = dict([(n.upper(), n) for n in session.aws_data_field_names])
+            if service_map.get(service) not in session.aws_data_field_names:
                 print('  Service not found. Please use the service name below.')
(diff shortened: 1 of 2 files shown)


pacu/main.py
```diff
@@ -685,9 +685,13 @@ def parse_data_command(self, command: List[str]) -> None:
             self.print('\nSession data:')
             session.print_all_data_in_session()
         else:
-            service = command[1].upper()
+            service = command[1]
+            if service.upper() in session.aws_data_field_names:
+                service = service.upper()
+
             if service not in session.aws_data_field_names:
```
RyanJarv, line 692: Hmm, I'm still having issues here. Does this work for you?

```suggestion
            if service.upper() not in [svc.upper() for svc in session.aws_data_field_names]:
```

RyanJarv (changes requested):

RyanJarv:
Welp, probably need some testing in general. Until then though left a suggestion that I believe should work in all cases.

pacu/main.py
naikordian: service like `CodeBuild` must use case sensitive some aren't. Like`IAM` or `EC2` both works fine with case insensitive.
```
APIGateway      CloudTrail      CloudWatch      CodeBuild       Config  DataPipeline    
DynamoDB        EC2     ECS     Glue    GuardDuty       IAM     Inspector       Lambda  
Lightsail       S3      SecretsManager   Shield  SSM     VPC     WAF     Account 
AccountSpend    Route53 RDS
```

naikordian:
```python
if service.upper() not in [svc.upper() for svc in session.aws_data_field_names]:
```
`codebuild` will pass the check but will fail in `getattr(session, service)` cause did't have  `codebuild` service name in `session.aws_data_field_names` only have service name `CodeBuild`

naikordian:
If you really want it to work in all cases you must rename fields in `session.aws_data_field_names` to uppercase.
 I can't think of a better way to convert some like `apigateway` to `APIGateway` case like this should be case sensitive

pacu/main.py
RyanJarv: Yeah so was hoping to work around that by converting everything to uppercase, looks like we need to do it for retrieving the data as well. Will make another suggestion here.

RyanJarv:
@naikordian made a second attempt at this if you want to try that (wouldn't let me make this suggestion this time, so committed to the bug/data_command branch instead).

naikordian:
@RyanJarv LGTM

RyanJarv (approved):
lgtm. FYI fixed the linting that was failing before and added some testing for this function.

PR #277, opened 2021-07-03
Improve Pacu's update function

naikordian:
Whenever open Pacu `check_for_udpate` take 5-10 seconds. This PR reduces checking update from every time to every week and use local cache update information


RyanJarv:
This is great, love the tests!

Made a few minor changes, simplifying the checks slightly (I hope at least), and formatting. Also regenerated the lock and export files (difficult to audit changes there, so this is a bit of a work around until this is done in CI). Definitely let me know if any of these changes aren't what you intended though.

RyanJarv (approved):

PR #279, opened 2021-07-10
Fix for users.json files not getting populated

sethsec:
I noticed that `run iam__privesc_scan --offline` was only reporting on roles and not users.  Digging into it, I realized that user json files were not getting written to  `~.local/share/pacu/test/downloads/confirmed_permissions/` when `iam__enum_permissions` is run with the `--all-users` flag, while the roles.json files were getting written.  

It looks like the only difference is this `session.name`. Once I removed it, everything worked as expected.  I ran `run iam__enum_permissions --all-users` and the user json files were written at which point `run iam__privesc_scan --offline` reported on both users and roles.  Easy fix for my use case, but I haven't done any other testing to see if this breaks something else.

### DIFF
--- a/pacu/modules/iam__enum_permissions/main.py
+++ b/pacu/modules/iam__enum_permissions/main.py
@@ -458,7 +458,7 @@ def main(args, pacu_main: 'Main'):
                         deny_permissions=user['Permissions']['Deny']
                     )
                 else:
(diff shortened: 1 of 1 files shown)


sethsec:
@RyanJarv - Let me know if you'd like me to create an issue for this, or if this PR is sufficient.

RyanJarv (approved):
Looks good to me, appreciate the PR!

PR #282, opened 2021-07-13
AWS Transfer family enumeration

juliosmelo:
Enumerates AWS Transfer Family SFTP/FTP and FTPS servers


RyanJarv:
On this PR it looks like the changes are primarily this commit: https://github.com/RhinoSecurityLabs/pacu/pull/282/commits/588fd57eb0f23425b13e380d72159b3758f4582b

Any reason for changing these names? Feel like they make sense as is, so not sure if I'm missing something.

Issue #284, opened 2021-07-19
Lambda backdoor new roles module file not found issue


dacelbot:
Hi all, thanks for your great work on Pacu! 
I'm having an issue with the `lambda__backdoor_new_roles` module, where if I run either the pip or docker versions of Pacu, I get the following error message:
`[2021-07-19 14:25:33] Pacu encountered an error while running the previous command. Check /root/.local/share/pacu/dockersesh/error_log.txt for technical details. [LOG LEVEL: MINIMAL]

    <class 'FileNotFoundError'>: [Errno 2] No such file or directory: './modules/lambda__backdoor_new_roles/lambda_function.py.bak'`

PR #285, opened 2021-07-28
fix pacu_dir not return '/home/user/.local/lib/python3.8/pacu'

naikordian:
If install pacu via pip pacu_dir() will return  `/home/user/.local/lib/python3.8/site-packages` 
this PR fix pacu_dir() to return `/home/user/.local/lib/python3.8/site-packages/pacu`

I found this when trying to fixes a bug on #284


RyanJarv (approved):
Thanks for the PR! Just added a test for the pacu_dir command but lgtm.

PR #287, opened 2021-07-31
fix lambda__backdoor_new_roles  path bugs

naikordian:
The module used relative path cause bugs when using Pacu from pip 

fix: #284


RyanJarv (approved):
Awesome thanks for the PR!

Added a note about the GetRole permissions requirement which I happened to notice when testing this, lgtm though.

PR #288, opened 2021-08-01
add a data command sub service filter

naikordian:
printing all data in the service is too much information. this PR helps you scope which sub-service you want to lookup.

```console
Pacu (session:test) > data ec2 subnets
```


RyanJarv (approved):
Awesome, this is a great change. Thanks for the PR!

PR #289, opened 2021-08-03
Feature/cfn__resource_injection

RyanJarv:
Module for injecting IAM admin roles into CloudFormation templates between upload and deploy.

Given an S3 bucket used for storing CloudFormation templates to be deployed this module will set up the S3 bucket notifications to trigger a lambda in another account when these templates are uploaded. This lambda will then inject an IAM admin role into the template, assuming this modification happens before the template is deployed, the user deploying is an admin, as well as deploys with the CAPABILITY_IAM permission (this more than likely the case) our IAM role will be deployed with the rest of the resources.

More info is included in the description of the module.


cmd-ctrl-freq (approved):
LGTM. Great work!

j0eblow (approved):
LGTM. Nicccce!

j0eblow (dismissed):
Nice.

RyanJarv:
Sorry for the last minute changes here. Should be good now.

j0eblow (approved):

PR #291, opened 2021-08-04
Release v1.1.0, add cfn_resource_injection testing

RyanJarv:
* Bump version to v1.1.0
* Adds the `pacu/modules/cfn__resource_injection/cfn__resource_injection_lambda/tests` tests to run during `make test`.
* Add moto to dev dependencies. 
* Dev requirements are exported without hashes. This was how it was done previously due to an error with colorama during install, thought this was fixed but that doesn't seem to be the case.
* Includes fixes for the dev Dockerfile.
  * The shell wasn't using the installed pip dependencies before.

### DIFF
--- a/pacu/modules/cfn__resource_injection/cfn__resource_injection_lambda/tests/test_app.py
+++ b/pacu/modules/cfn__resource_injection/cfn__resource_injection_lambda/tests/test_app.py
@@ -8,12 +8,6 @@ from chalice.test import Client
 from pacu.modules.cfn__resource_injection.cfn__resource_injection_lambda.app import app, add_role_dict, add_role_bytes, fetch, update
 
 
-def test_lambda_handler():
-    with Client(app) as client:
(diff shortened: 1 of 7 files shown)


pacu/modules/cfn__resource_injection/cfn__resource_injection_lambda/tests/test_app.py
```diff
@@ -164,8 +158,14 @@ def s3(aws_credentials):
 
 def test_update(s3, upload_json):
     bucket, key, body = upload_json[0], upload_json[1], upload_json[2]
```
cmd-ctrl-freq, line 160: Seems like we don't need to assign a value to `body` here.

cmd-ctrl-freq (commented):
I have some feedback that might be 'nit-picky'. It looks like there is an unnecessary assignment.

cmd-ctrl-freq (approved):
Looks good!

PR #293, opened 2021-08-05
Feature/migrations

RyanJarv:
This adds some basic support for migrations.

Currently if you add an modules for new services it will break Pacu on update because the DB schema has changed. The fix currently is to to delete and and recreate the database. This also is somewhat worse now with it using a common DB for all installs, can't just re-clone and have it work like normal again.


cmd-ctrl-freq (approved):
Nice work!

PR #299, opened 2021-08-18
Update cfn__resource_injection with region arg to fix exception

dacelbot:
Thanks for your work! I love Pacu. 
In testing the new module cfn__resource_injection, I found that if you specify a region other then us-east-1 for the bucket location, it caused an exception on line 353 of _add_lambda_permission_ in main.py for this module. This appeared to be caused by the fact that the attacker_sess variable didn't have a region specified, and therefore defaults to us-east-1, so in turn _add_lambda_permission_  only searches us-east-1 for a Lambda and one will not exist. 
I solved this for my use by adding a CLI flag _--attacker-region_ which the user specifies at runtime, and figured I'd contribute the change back. There's certainly other ways this issue could be solved, but here's one!


RyanJarv:
To keep this up to date I believe this change will fix the issue without adding an additional argument: https://github.com/RyanJarv/pacu/commit/6da90bf80092ae6cad1d9aab3bed59ce8fe435e3.

I still need to test this though.

PR #301, opened 2021-09-01
Add a zip_file util

naikordian:


RyanJarv:
Awesome, thank you for the PR!

Have been thinking we need to do something about extracting some of the lambda functionality out into common code. I may need to come back and review/merge this a bit later this week just FYI. Also like always definitely appreciate the tests here!


Related to this, it's likely worth noting that in the latest cloudformation_resource_injection module I actually ended up using chalice to build and deploy the lambda. I still want to do more evaluation if it makes sense to start going in that direction for managing lambdas, but because that's still somewhat undecided whether that makes sense as a general solution, and the fact it will take some time to migrate either way, any PR to clean up the existing lambda code will be useful.

As a note to my self I'll open another issue on an overview of adopting chalice more generally across Pacu, that way anyone can track the progress there or add to that discussion (and actually think we should start doing this regularly for more significant changes/additions like this in the future).

DaveYesland (approved):
Thanks for the pull request and addition to Pacu! sorry for the delay.

Issue #303, opened 2021-09-08
iam__backdoor_assume_role overwrites IAM boto client object when user arn is not specified


ishmandoo:
It looks like this line is looking for an IAM boto client object. However, `client` gets overwritten with an STS boto client on line 80 if a user arn is not specified.

https://github.com/RhinoSecurityLabs/pacu/blob/e95b3ff2898b4954fe61411b314414b33e1f5750/pacu/modules/iam__backdoor_assume_role/main.py#L105

The result is an error like this:
`<class 'AttributeError'>: 'STS' object has no attribute 'update_assume_role_policy'`

ishmandoo:
I just put up a PR proposing a fix: #304

DaveYesland:
Looks like this was merged and fixed

PR #304, opened 2021-09-11
Fix for IAM Boto client overwrite bug

ishmandoo:
This PR addresses https://github.com/RhinoSecurityLabs/pacu/issues/303. In `iam__backdoor_assume_role ` the IAM Boto client gets overwritten with an STS Boto client if a user ARN is not specified. I'm proposing a fix that just uses a different name for the STS client.


RyanJarv (approved):
Thank you! lgtm

PR #307, opened 2021-10-09
adding support of China AWS

zhujian0805:


phefley (commented):
I can't pull this code in for you, but I can confirm that it allowed the update_regions task to bring in the Chinese regions as expected! Thank you!

PR #308, opened 2021-11-07
Remove .bak files and format code

notdodo:


RyanJarv:
This might be difficult to merge as is. Currently I believe these bak files are used, so at least we'll want to remove those changes from the PR. One example of these [files](https://github.com/RhinoSecurityLabs/pacu/blob/e95b3ff2898b4954fe61411b314414b33e1f5750/pacu/modules/lambda__backdoor_new_sec_groups/lambda_function.py.bak) being used is [here](https://github.com/RhinoSecurityLabs/pacu/blob/e95b3ff2898b4954fe61411b314414b33e1f5750/pacu/modules/lambda__backdoor_new_sec_groups/main.py#L156). Using .bak as a file extension in this way is a bit strange but I don't think this makes sense to change at the moment, likely will take more digging to figure out a better way to do this.

For merging the formatting changes I think we would need a way to reproduce and rerun the code formatting as needed. So black would need to be included in the dev dependencies and the formatting command likely makes sense to include as a makefile task.

RyanJarv:
If you want to make the changes suggested in the last comment we can reconsider merging this PR, we may need to come back to this a bit later otherwise.

Besides that though I thing these changes are a good suggestion, we do need to improve the formatting and (maybe to a lesser degree) consider if we can handle the .bak files in a way that makes more sense.

PR #313, opened 2022-01-19
Enhancement/download ec2 public ip

Gfuen:
Enhancement code has been added to the ec2__enum module. Enhancement has been tested and file(s) show up according to region. Code has been tested according to flake8 and fixed mistakes accordingly. Please let me know if anything needs to be changed in the code or anything. Thank you.


RyanJarv (approved):
Thanks for the PR! I made a small change [here](https://github.com/RhinoSecurityLabs/pacu/pull/313/files#diff-074a6a4fdc4b6324834cf1bbc7467f3461a01d36ab7d25aa10a2093dbdd5e2b2R247) that may causes issues when an instance doesn't have a public IP address but besides that looks good.

Gfuen:
Sounds good and sorry about the small forgotten edge case!

PR #318, opened 2022-04-27
Extend cli: add new session and set AWS keys

agroyz:
Extend cli: add new session and set AWS keys


DaveYesland (approved):
Thanks for the pull request and addition to Pacu! sorry for the delay.

PR #319, opened 2022-06-06
cfn__resource_injection: Wrongly checking session id of victim as current session

jyenduri-uptycs:
Currently for the module cfn__resource_injection takes two input

1. s3-access-key : Victims key
2. attacker-key: Attackers key

In the code we are checking if current session is victims key or not and throwing an error. Victims key and Attacker key cannot have same pacu session at same time, we should just check if victims key exists in pacu database and use that as needed.   Function `get_aws_key_by_alias_from_db` checks if give key alias exists in db or not. 

```
Pacu (attack:attack) > run cfn__resource_injection --attacker-key attack --s3-access-key victim 

```

### DIFF
--- a/pacu/main.py
+++ b/pacu/main.py
@@ -608,11 +608,11 @@ class Main:
          PacuSession from the database, or None if no AWSKey
         with the supplied alias exists. If more than one key with the alias
         exists for the active session, an exception will be raised. """
-        session = self.get_active_session()
+        # session = self.get_active_session()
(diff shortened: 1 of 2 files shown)


jyenduri-uptycs:
@RyanJarv  Please review

DaveYesland (approved):
Thanks for the pull request and addition to Pacu! sorry for the delay in getting this merged.

Issue #286, opened 2021-07-29
Feature request: switch aws cli profile from within pacu


sfellows-tii:
Currently, when invoking aws cli commands (e.g. 'aws configure list'), the aws cli profile is the same as when pacu was started. If you want to switch to another aws cli profile, you have to quit pacu, change the aws cli profile, then restart pacu. It would be nice to be able to do this from within pacu or be able to invoke a shell to do this.

I am using the swap_keys functionality in pacu to run modules, etc. using different credentials. It is only when invoking aws commands that there is a mismatch between profiles.

PR #322, opened 2022-08-24
Using session keys as profiles in internal aws command

Betep0k:
Hi! Feature allows to use session keys as profiles in an internal aws command. So if you have active keys, they will be used as default profile when you execute aws command. Also if user has selected 1 region it also will be used as default region in aws command. 

Users can use system aws configs by adding `--profile` flag. For example, `aws --profile default ...` will use default system config. 

Code creates temporary config files and changes environment variables before the run of aws command. As a result aws command uses temporary files as "credentials" config. After execution it deletes all temporary files.

### DIFF
--- a/pacu/main.py
+++ b/pacu/main.py
@@ -609,7 +609,40 @@ class Main:
         command = command.strip()
 
         if command.split(' ')[0] == 'aws':
-            self.run_aws_cli_command(command)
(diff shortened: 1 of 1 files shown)


DaveYesland (approved):
Thanks for the pull request and addition to Pacu! sorry for the delay in getting this merged. Resolves #286

PR #328, opened 2023-01-04
Updated the iam__detect_honeytokens module to use SNS Publish

BenFromKC:
Updated the iam__detect_honeytokens module to use SNS Publish for unlogged detection goodness (SDB ListDomains is now logged to CloudTrail). 

Shout out to Nick Frichette for finding the SNS Publish option.


DaveYesland (approved):
Thanks for the pull request and update to Pacu!

PR #347, opened 2023-04-27
Add privesc method info feature

DaveYesland:
This just adds an enhancement/feature to the iam__privesc_scan module so that you can list privesc methods and print out information about a particular privesc method.


jdearmas (approved):
This pull requests adds two command-line arguments to the iam__privesc_scan module:

-  '--method-info'
- '--method-list'

PR #354, opened 2023-05-31
Fix no loop break when enter for key alias

DaveYesland:
Fixes a bug where the the loop when entering a key alias never breaks if the user just presses enter to keep the same value.


jdearmas (approved):
This commit allows a user to provide an empty string in the 'set_keys()' pacu menu in order to exit the menu.

PR #356, opened 2023-05-31
Change session_token to be none when keys are set without one fixes #355

### LINKED ISSUE #355: AWS API calls fail when importing user keys from profile after setting temp keys with session token
If you set temporary keys including a session token in a session and then import standard user keys from a profile without a session token all calls will then fails because the session token is never cleared during the set_keys process so it is still sent when authenticating.

```
    <class 'botocore.exceptions.ClientError'>: An error occurred (InvalidClientTokenId) when calling the GetCallerIdentity operation: The security token included in the request is invalid
```

Steps to reproduce:

1. Set keys to temporary creds which include a session token
2. run `set_keys` and enter through the menu to ensure all values are set
3. run `import_keys` and import normal user keys without a session token
4. run `set_keys` again to see the session token is still set
5. run `run iam__enum_permissions` to see call fails because session token is still set

DaveYesland:
Fixes #355

### DIFF
--- a/pacu/main.py
+++ b/pacu/main.py
@@ -1334,9 +1334,7 @@ aws_secret_access_key = {}
         if key_alias is None:
             new_value = self.input('Session token (Optional - for temp AWS keys only) [{}]: '.format(session.session_token))
         else:
-            new_value = session_token
-            if new_value is None:
-                new_value = 'c'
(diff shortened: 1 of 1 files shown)


jdearmas (approved):
I have tested this pull request and can affirm that it works as intended.

PR #357, opened 2023-05-31
Enhancement/add cli option to import keys

DaveYesland:
This adds the option to import keys into a session when running a Pacu module from the CLI.

For example:
`poetry run pacu --module-name iam__enum_permissions --exec --import-keys SOME_AWS_PROFILE --session SOME_SESSION`


jdearmas (approved):
I have tested this pull request and can affirm that it works as intended.

PR #363, opened 2023-07-05
Cognito Enumeration and Attack Modules

davidkutz-marks:
This pull request will add the cognito__enum and cognito__attack modules to Pacu.


davidkutz-marks:
Dependencies for attack module:
base64
pprint
webbrowser
qrcode
pycognito
re

PR #364, opened 2023-07-06
New Cognito modules

davidkutz-marks:
This PR adds two new Cognito modules, one for enumeration and one for attack.


DaveYesland (commented):
Everything looks good to me. I was able to run this with no issues as well, worked as intended. Will wait for @jdearmas to give it a final approval though.

PR #383, opened 2024-01-04
Remove old install method

DaveYesland:
This cleans up an removes the old install.sh which is no longer the proper install method.


jdearmas:
I got this error when installing using `pip install -U pacu`
image

davidkutz-marks (approved):

PR #421, opened 2024-05-20
Enhance iam__enum_permissions w/ unconfirmed and counts

h00die:
This PR modifies `iam__enum_permissions` to display unconfirmed permissions, and counts of each.

## Pre

To start, we have a new user and key entered in to the db.
image
When running `iam__enum_permissions` for this (limited) user, we get back the following output.
image
When looking at this output, it looks like the module failed. We have a `FAILURE`, a `FAILURE`, and while it says 'confirmed permissions for ...', nothing is listed like a user would expect, so I assume no permissions were found BUT the query didn't fail.  We move on down to the summary, and we see 0 and 0. So we must not have been able to enumerate.
However, we run a `whoami` and we see we *did* get permissions:
image

The problem is that these are **unconfirmed** permissions.  However, they were found so the user should be notified of something.

## Post

We now keep track of how many unconfirmed items were present, and a count of the actual permissions. This allows us to give the user better output so they know data was found, which matches what they'll see in `whoami`.

image

### DIFF
--- a/pacu/modules/iam__enum_permissions/main.py
+++ b/pacu/modules/iam__enum_permissions/main.py
@@ -559,7 +559,7 @@ def summary(data, pacu_main):
                                                                                   count_max_len=count_max_len)
     else:
         out += "{:>{count_max_len}} Unconfirmed permissions for {} user(s).\n".format(data["users_unconfirmed_perm_count"],
-                                                                                  data["single_user"],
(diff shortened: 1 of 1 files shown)


h00die:
pulling back to draft to fix the R justification issue

DaveYesland (approved):

PR #429, opened 2024-05-24
fix iam__decode_accesskey_id category

h00die:
categories are always UPPERCASE. `iam__decode_accesskey_id` seems to be a one off, which is causing it to get its own menu heading when `ls` is run, instead of being with the rest.

## Pre

image

## Post

(second from the bottom)
image


DaveYesland (approved):

DaveYesland (approved):

h00die:
hold off on the release, should have a new mq enum module coming shortly!

DaveYesland:
Thanks for the heads up.

PR #431, opened 2024-05-24
New module: mq__enum

6a6f656c:
Enumerates MQ Instances and stores information into json file.


h00die:
I'm doing final testing on this now, will have screenshots and other information shortly.

h00die:
## Running against a specific region
image

For the stored data, it's just a subset of the data that we thought would be of value to a pentester.
image

## all regions with 1 broker & 1 user

image

image

## mutliple brokers and users

image

image

h00die:
The only other improvement we thought of (other than adding more useful fields into the saved data) would be saving this to `data` as well. Thoughts on that?

DaveYesland:
> The only other improvement we thought of (other than adding more useful fields into the saved data) would be saving this to `data` as well. Thoughts on that?

Yeah agreed, seems like this data should go into the DB instead of just a file since it is just structed data from a service.

pacu/modules/mq__enum/main.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+        except Exception as error:
+            print(
+                "Unable to list brokers; Check credentials or No brokers are available. Error: {}".format(
+                    error
+                )
+            )
+            continue
+        print("Found {} brokers".format(len(response["BrokerSummaries"])))
```
h00die, line 77: Add 2 spaces to the start of this print string

6a6f656c:
All fixed...ready for review.

pacu/modules/mq__enum/main.py
```diff
@@ -0,0 +1,136 @@
...
+from pacu.core.lib import downloads_dir
+from pacu.core.lib import strip_lines
+from pacu import Main
+from copy import deepcopy
+
+module_info = {
+    "name": "mq__enum",
+    "author": "6a6f656c & h00die of nDepth Security",
+    "category": "ENUM",
+    "one_liner": "Listo and describe brokers",
```
DaveYesland, line 15:
```suggestion
    "one_liner": "List and describe brokers",
```

pacu/modules/mq__enum/main.py
```diff
@@ -0,0 +1,136 @@
...
+            continue
+
+        # Prepare output file to store MQ data
+        now = time.time()
+        outfile_path = str(downloads_dir() / f"mq_enum_{now}.json")
+
+        try:
+            response = client.list_brokers(
+                MaxResults=100,
+            )
```
DaveYesland, line 73: Maybe use a paginator instead of maxresults to get everything
```suggestion
             paginator = client.get_paginator('list_brokers')
            # Create a PageIterator from the paginator
            page_iterator = paginator.paginate()
```

pacu/modules/mq__enum/main.py
```diff
@@ -0,0 +1,136 @@
...
+
+        except Exception as error:
+            print(
+                "Unable to list brokers; Check credentials or No brokers are available. Error: {}".format(
+                    error
+                )
+            )
+            continue
+        print("  Found {} brokers".format(len(response["BrokerSummaries"])))
+        for broker in response["BrokerSummaries"]:
```
DaveYesland, line 83: If using the paginator
```suggestion
        for response in page_iterator:
                for broker in response["BrokerSummaries"]:
```

pacu/modules/mq__enum/main.py
```diff
@@ -0,0 +1,136 @@
...
+            broker_details = client.describe_broker(BrokerId=broker["BrokerId"])
```
DaveYesland, line 84: I noticed if a broker is in the state `CREATION_IN_PROGRESS` then it errors out so this should be checked for:
```suggestion
            if broker["BrokerState"] == "CREATION_IN_PROGRESS":
                    continue
            broker_details = client.describe_broker(BrokerId=broker["BrokerId"])
```

pacu/modules/mq__enum/main.py
DaveYesland, line 83: Probably should do `broker_id = broker["BrokerId"]` then use `broker_id` in all the references below just to make it a little cleaner.

pacu/modules/mq__enum/main.py
6a6f656c: page_iterator stays dimmed out as if unused

6a6f656c:
Ready for review

DaveYesland (approved):

DaveYesland (approved):

PR #433, opened 2024-05-29
alias use to run/exec

h00die:
Make pacu more friendly to users of other frameworks (metasploit) by aliasing `use` to `run` and `exec`


DaveYesland (approved):
Looks good and worked with my testing.

PR #436, opened 2024-05-30
Improve tests

EduardSchwarzkopf:
This PR should improve the tests for the modules. I'm going to create a structure that allows to create of tests in each module. Will see if I can provide some sort of template for the mock tests with moto.


EduardSchwarzkopf:
moto is not sufficient enough. It generates predefined responses, which is ok for small tests, but not when you want to try out different things. E.g. for the `cognto__attack` the user is already verified, so you cannot check for the verification input.
I'm going to try out `localstack` and see if this will give a more suitable testing solution

EduardSchwarzkopf:
This PR is the basis for integrating further module tests more easily. I have provided an example for `cognito__attack`. I will add more here, but I prefer to do this in a separate PR

DaveYesland:
This looks great! thank you for all these additions, testing is something that certainly needs work in Pacu. We will start reviewing this when time allows and will get it merged.

DaveYesland (approved):
All looks good to me, tests all pass locally for me.

PR #437, opened 2024-05-31
Fixed the apigateway__enum module

TeneBrae93:
- If the user has "GET" permissions on API Gateways but not the API Keys, the module will still get all the paths/methods/resources rather than failing

- There was no error handling on "all regions" so it would fail. Added error handling so the user can check all regions without issues

- Previously it saved the data to a standalone file, updated this so it stores to the Pacu DB like other modules

### DIFF
--- a/pacu/modules/apigateway__enum/main.py
+++ b/pacu/modules/apigateway__enum/main.py
@@ -1,15 +1,10 @@
 #!/usr/bin/env python3
 
 import argparse
-import os
-from pathlib import Path
 import pprint
 import json
-import time
 from copy import deepcopy
 
-from pacu.core.lib import downloads_dir
-from pacu.core.lib import strip_lines
 from pacu import Main
 
 module_info = {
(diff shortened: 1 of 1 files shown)


pacu/modules/apigateway__enum/main.py
```diff
@@ -141,43 +151,50 @@ def parse_method(base_url, method, path, stages):
...
-def main(args, pacu):
+def main(args, pacu_main: "Main"):
     """Main module function, called from Pacu"""
-    print = pacu.print
-    session = pacu.get_active_session()
+    session = pacu_main.get_active_session()
+    print = pacu_main.print
     args = parser.parse_args(args)
 
     outfile_path = downloads_dir()/'apigateway'
```
DaveYesland, line 160: We should remove all file output stuff if this data is now saved in the DB

TeneBrae93: Stand by, I think I missed something on the database. Digging into this now and will also fix this

TeneBrae93: @DaveYesland -- This should all be fixed now

DaveYesland (approved):

PR #439, opened 2024-06-04
Adds a debug command to view error log

DaveYesland:
This just adds a debug command so that you can easily dump the error log from within the Pacu shell.


TeneBrae93 (approved):
Confirmed everything works as intended.

PR #443, opened 2024-06-21
Module to extract service account tokens of running pods in EKS

guragainroshan0:
Module to retrieve service account tokens of running pods in a EKS cluster. Credentials retrieved from SSRF could be role of a node. This could be exploited to extract service account tokens of pods in the cluster and exploit it further. 
This module requests EKS authentication token for a node and uses it to get service account token of pods running in the cluster.


DaveYesland:
Starting to take a look at this and test it. Thanks for the contribution!

I think something more specific for the module name might be better, maybe: `eks__collect_tokens` or `eks__collect_service_tokens` something to make it clear what it does at a glance.

guragainroshan0:
Thanks for the review.  

The module name has been modified to eks__collect_tokens.

chebuya (approved):
Hey @guragainroshan0, thanks for contributing!  I have reviewed and tested the module, all is looking great so will be merging it soon

PR #447, opened 2024-08-01
Add directory services enumeration

chebuya:
Add module for directory services enumeration


nobodynate (changes requested):
Working when `--region` is correct.
Getting error when region is incorrect, which causes issues when region is not specified and correct.
Maybe adding error handling for the case when there is not a deployment in the specified region will help

chebuya:
Fixed issue with all regions, should be good now!

> Working when `--region` is correct. Getting error when region is incorrect, which causes issues when region is not specified and correct. Maybe adding error handling for the case when there is not a deployment in the specified region will help

nobodynate (approved):
working. thanks!

PR #448, opened 2024-08-12
Update secrets_enum to list found secrets

h00die:
When a user has permission to list secrets, but not read them, the output isn't very helpful:
image

Lots of failures, but no idea what the failures are for. Yes access is denied, but to what?

This PR adds a few print statements to give the user a better idea of what is being probed.
image


chebuya (approved):
Hey @h00die, thanks for contributing.  I have reviewed it and will be merging it soon.

PR #460, opened 2024-12-31
Add command to delete AWS keys

luisfontes19:
Personally I prefer to delete some keys as soon as I don't need them anymore just for peace of mind, right now there's no options for that and you need to go directly to the DB or delete the current session 😞

This PR adds a simple `delete_keys` command to delete chosen keys from the current session


DaveYesland (approved):
Thanks for this addition. Just added the `delete_keys` to the main help menu but looks good to me.

PR #471, opened 2025-03-12
[WIP] - Fix ebs__download_snapshots region issue

TeneBrae93:
## Summary

This PR updates the Pacu module for downloading EBS snapshots to explicitly pass the region to dsnap’s `LocalSnapshot` constructor. By doing so, dsnap is forced to create its boto3 session using the correct AWS region (e.g., "us-west-2"), rather than defaulting to "us-east-1". This prevents errors like `ResourceNotFoundException` when the snapshot exists in a non-default region.

## Important
The Dsnap PR needs to also be approved for this to work properly - https://github.com/RhinoSecurityLabs/dsnap/pull/29


TeneBrae93:
I believe the errors from the build process are due to it relying on the old version of dsnap; the dsnap PR will need to be approved first

DaveYesland (approved):

PR #479, opened 2025-05-16
Fix backspace rendering issue by setting readline mode

WireHawkSecurity:
When entering input in Pacu, backspace displayed ^H instead of deleting characters. This was due to readline not having a defined editing mode. Adding readline.parse_and_bind('set editing-mode emacs') ensures backspace functions as expected in most terminals.


WireHawkSecurity:
made an update to the comment to proper format

